PR #47, opened 2026-03-09
feat(frontend): backend-persisted drafts center + composer stability fixes

ALI-RUBASS:
## What
- add backend drafts API client (`/service/users/drafts`) and wire exports/endpoints
- move composer draft persistence from old/local flow to backend draft records
- add new Drafts Center page and route/navigation entry
- add draft metadata/index support for chat/thread restore behavior
- add i18n labels for Drafts navigation and page

## Fixes included
- prevent chat input lock when draft exists and plaintext draft is loaded
- hide internal draft-index record from Drafts Center list/details
- place caret at end of loaded draft text when reopening chat
- align drafts list fetch to safe page size and avoid error-page redirect behavior in draft API handling

## Verification
- linted changed files (`src/api/apiUserDrafts.ts`, `src/components/draftsCenter/DraftsCenter.tsx`, `src/components/messageSubmitInterface/useDraftMessage.tsx`)
- deployed frontend to dev and verified rollout healthy
- validated that drafts page no longer shows internal JSON entry
- validated that reopening a draft places cursor at message end

## Scope
- Repository: `ORISO-Frontend`
- Target branch: `dev`
- Source branch: `ds-drafts-system`

### DIFF
--- a/src/components/draftsCenter/DraftsCenter.tsx
+++ b/src/components/draftsCenter/DraftsCenter.tsx
@@ -1,0 +1,172 @@
+import * as React from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useHistory } from 'react-router-dom';
+import { useTranslation } from 'react-i18next';
+import { apiDeleteUserDraft, apiGetUserDrafts, IUserDraftItem } from '../../api';
+import { REMOTE_DRAFT_INDEX_SCOPE } from '../../services/draftStore';
+import './draftsCenter.styles';
+
+const formatRelativeTime = (timestamp?: string | null) => {
+	if (!timestamp) {
+		return '';
+	}
+	const diffMs = Date.now() - new Date(timestamp).getTime();
+	const diffMin = Math.max(0, Math.floor(diffMs / (1000 * 60)));
+	if (diffMin < 1) return 'now';
+	if (diffMin < 60) return `${diffMin}m`;
+	const diffHours = Math.floor(diffMin / 60);
+	if (diffHours < 24) return `${diffHours}h`;
+	const diffDays = Math.floor(diffHours / 24);
+	return `${diffDays}d`;
+};
+
+export const DraftsCenter = () => {
+	const { t: translate } = useTranslation();
+	const history = useHistory();
+	const [selectedDraftKey, setSelectedDraftKey] = useState<string | null>(null);
+	const [refreshToken, setRefreshToken] = useState(0);
+	const [drafts, setDrafts] = useState<IUserDraftItem[]>([]);
+
+	useEffect(() => {
+		let isMounted = true;
+		const loadRemoteDrafts = async () => {
+			const response = await apiGetUserDrafts(0, 200).catch(() => null);
+			if (!isMounted) {
+				return;
+			}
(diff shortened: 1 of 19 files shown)


src/components/messageSubmitInterface/useDraftMessage.tsx
```diff
@@ -16,86 +17,164 @@ import {
...
+			setMessage(messageRes.text);
 			setLoaded(true);
 			return;
 		}
 
+		if (!ready) {
+			return;
+		}
+
 		decryptText(
```
shazia-k, line 169: Issue: Encrypted drafts are saved, but restore never decrypts them. The hook saves encrypted text, then calls decryptText(..., false, 'enc.'), which bypasses decryption.

Suggestion: Persist an encryption marker like the old t === 'e2e', or infer encryption from the enc. prefix before calling decryptText.

src/components/messageSubmitInterface/messageSubmitInterfaceComponent.tsx
```diff
@@ -304,8 +305,43 @@ export const MessageSubmitInterfaceComponent = ({
...
+				: activeSession?.item?.topic;
+		return (
+			contact?.displayName ||
+			contact?.username ||
+			topicName ||
+			null
+		);
+	}, [activeSession]);
+
+	const {
```
shazia-k, line 334: Issue: The cleanup save now runs on every message change, and there is no timeout cancellation before draft deletion. After send/delete, a pending debounce can re-save the draft that was just cleared.

Suggestion:
Restore the old unmount-only cleanup guard and clear the pending draft timeout before delete/send flows.

src/components/draftsCenter/DraftsCenter.tsx
```diff
@@ -0,0 +1,172 @@
...
+								<div className="draftsCenter__listItemHeader">
+									<span className="draftsCenter__listItemTitle">
+										{entry.title ||
+											translate('drafts.center.untitledChat', 'Chat')}
+									</span>
+									<span className="draftsCenter__listItemTime">
+										{formatRelativeTime(entry.updatedAt)}
+									</span>
+								</div>
+								<p className="draftsCenter__listItemText">{entry.text}</p>
```
shazia-k, line 124: Drafts Center renders raw entry.text with no encryption metadata or decrypt path. Secure drafts will show stored encrypted payloads in the new /drafts view.

Suggestion is Store a safe preview separately, or show an Encrypted draft placeholder until the draft is opened in its original chat context.

shazia-k (commented):
Need to implement suggested changes.

src/components/messageSubmitInterface/useDraftMessage.tsx
ALI-RUBASS: The encryption is manged through the matrix server (synapse) that is why the local decryption (frontend) is not called. The text strings are fetched directly moving through the synapse.

src/components/draftsCenter/DraftsCenter.tsx
ALI-RUBASS: local decryption is not used rather matrix synapse decryption is used.

PR #57, opened 2026-03-29
Stabilize TipTap composer modal, alignment, audience selector, and sidebar focus UX

ALI-RUBASS:
## Summary
- Rework the expanded composer presentation into a stable centered modal flow and tighten related wrapper layering behavior.
- Normalize TipTap rich-text alignment defaults and rendering so new input, restored drafts, and displayed messages consistently preserve left alignment unless explicitly changed.
- Fix audience selector visibility and options behavior for participant counts, and refine session list card/focus styling for clearer active-item separation and keyboard navigation feedback.

## Test plan
- [ ] Open a chat and verify normal composer opens and accepts input immediately after refresh.
- [ ] Click maximize and verify expanded composer appears centered, with usable dimensions and no sidebar overlap.
- [ ] Type new content and confirm left alignment is default; switch alignments and verify toggles behave correctly.
- [ ] Reload and validate audience selector only appears in multi-participant chats and includes `Send to all` where expected.
- [ ] Navigate session list with keyboard and confirm active/focus visuals and adjacent card rounding behavior are clear.

Made with [Cursor](https://cursor.com)


src/components/messageSubmitInterface/TipTapComposer.tsx
```diff
@@ -276,12 +304,20 @@ export const TipTapComposer = forwardRef<TipTapComposerRef, TipTapComposerProps>
 			if (!editor) {
 				return;
 			}
+			const normalizedValue = (value || '')
```
shazia-k, line 307: Issue:
On every value sync: content is passed through normalizedValue that rewrites right alignment to left, then after setContent editor.commands.setTextAlign('left') runs unconditionally. That fights the new alignment feature and can reset or distort center/right alignment whenever parent state re-syncs.
Suggestion:
Drop the unconditional setTextAlign('left') after setContent (only set default alignment for truly empty content). Do not strip text-align: right / data-text-align="right" unless there is a documented, scoped reason (e.g. RTL-only).

src/components/messageSubmitInterface/messageSubmitInterfaceComponent.tsx
```diff
@@ -407,12 +519,28 @@ const AudiencePersonIcon = () => (
...
+	const [isExpandedComposer, setIsExpandedComposer] = useState(false);
 	const [isAudienceMenuOpen, setIsAudienceMenuOpen] = useState(false);
 	const [audienceOptions, setAudienceOptions] = useState<
 		Array<{ value: string; label: string }>
 	>([{ value: '__all__', label: 'ALL' }]);
 	const [selectedAudienceValue, setSelectedAudienceValue] =
 		useState<string>('__all__');
+	const [audienceRefreshTick, setAudienceRefreshTick] = useState(0);
+
+	const normalizeInitialAlignment = useCallback((rawValue: string) => {
```
shazia-k, line 536: Issue: normalizeInitialAlignment applies the same right→left rewrites when loading drafts. Users can lose saved right-aligned content on load.
Suggestion:
Align with product: either remove this normalization or limit it (e.g. only when forcing LTR UI).

src/components/messageSubmitInterface/messageSubmitInterface.styles.scss
```diff
@@ -69,6 +119,25 @@ $inputBoxShadow: inset 0 2px 0 0 #ffffff;
...
+				transition: all 240ms ease;
+
+				&--expanded {
+					position: fixed !important;
+					left: 50vw !important;
+					top: 50vh !important;
+					right: auto !important;
+					bottom: auto !important;
+					inset: auto !important;
+					margin-top: -850px;
```
shazia-k, line 131: Expanded composer uses margin-top: -850px with fixed positioning — fragile across viewports/DPI/zoom and can place the panel off-screen or inconsistently.
**Suggestion:** 
Center with transform: translate(-50%, -50%) (and sensible max-height/max-width) instead of a large negative margin.

src/components/sessionsListItem/SessionListItemComponent.tsx
```diff
@@ -561,6 +567,7 @@ export const SessionListItemComponent = ({
 					ref={itemRef}
 					tabIndex={index === 0 ? 0 : -1}
 					role="tab"
+					aria-selected={isChatActive}
```
shazia-k, line 570: Issue: Items use role="tab" and now aria-selected, but the scroll list is not wrapped in role="tablist" (unlike the real tab strip higher in SessionsList.tsx). Screen readers get an invalid tab pattern.
Suggestion:
Use role="list" / role="listitem", or role="button" with aria-current for the active chat, or wrap the list in a proper tablist only if keyboard model matches tabs.

src/components/messageSubmitInterface/messageSubmitInterfaceComponent.tsx
```diff
@@ -2105,6 +2224,11 @@ const AudiencePersonIcon = () => (
 		composerRef.current?.runAction(action);
 	}, []);
 
+	const isToolbarActionSelected = useCallback(
```
shazia-k, line 2227: Issue:
isToolbarActionSelected depends on [composerText] but reads live state via composerRef.current?.isActionActive. Selection/formatting can change without composerText changing, so toolbar “selected” states can be stale.
Suggestion:
Subscribe to TipTap onSelectionUpdate/onUpdate and bump state, or pass selection revision into the callback deps.

shazia-k (commented):
@ALI-RUBASS Comments added to implement

PR #58, opened 2026-03-31
feat(ui): align search bar filters and new-chat picker UI

ALI-RUBASS:
## Summary
- Reworked sessions search interaction and toolbar behavior to match requested Figma flow.
- Added open/close behavior so the search panel no longer stays over chat cards.
- Ensured outside-click closes the search panel and restores list visibility.
- Refined search reopen behavior when text already exists in the field.
- Updated search panel visual structure and icon states for improved UX consistency.
- Reworked new-chat participant picker to a dedicated custom interaction pattern.
- Added separate rounded cards for search, current participants, and participant list sections.
- Added selectable participant pills with remove controls in Current Participant section.
- Increased participant list and picker typography to match field readability.
- Matched participant field sizing/geometry to the top form controls.
- Added plus-button animated state transition (default light rounded to active gray rounded-square).
- Removed the redundant `New chat` card from sessions list create-chat route.
- Improved mobile chat tag visibility so labels like `Family counseling` are less truncated.

## Notes
- Drafts flow was checked: selecting/sending a draft staying in panel is intended behavior.

## Test plan
- [x] Verify search panel closes on outside click and no longer hides chat cards persistently.
- [x] Verify search panel can reopen correctly with existing text.
- [x] Verify participant picker cards render separately with rounded containers.
- [x] Verify participant select/deselect updates pills and list state.
- [x] Verify plus-button transition state in closed/open modes.
- [x] Verify `New chat` list card is removed from the create-chat route.
- [x] Verify mobile tag readability for long topic labels.


src/components/session/session.styles.scss
```diff
@@ -1918,4 +1926,53 @@ $session-scroll-to-bottom-radius: $button-border-radius !default;
...
 	}
+}
+
+/* Target the specific scrollable containers (like your list and chat detail) */
+.contentWrapper__list, 
+.contentWrapper__detail,
+.session__wrapper {
+  /* Firefox support */
+  scrollbar-width: thin;
+  scrollbar-color: transparent transparent;
```
shazia-k: Issue: The new scrollbar styling hides the scrollbar for .session__wrapper, but the hover/reveal rules do not include .session__wrapper:hover. In Firefox, the chat pane scrollbar remains invisible at all times.
Suggestion : Add .session__wrapper:hover to the reveal rules, or avoid hiding the thread scrollbar by default.

src/components/sessionsList/ResizableHandle.tsx
```diff
@@ -78,7 +78,7 @@ export const ResizableHandle: React.FC<ResizableHandleProps> = ({
 				bottom: 0,
 				width: '1px', // Wider for easier grabbing
 				cursor: 'col-resize',
-				backgroundColor: isDragging ? 'rgba(80, 80, 80, 0.2)' : 'rgba(80, 80, 80, 0.1)',
+				backgroundColor: isDragging ? 'rgba(0, 0, 0, 0)' : 'rgba(0, 0, 0, 0)',
```
shazia-k: Issue: The resize handle is fully transparent in both states. Since the hit area is only 1px, the resizer becomes nearly impossible to discover unless the user hovers exactly over it.
Suggestion: Keep a faint visible handle by default, or increase the hit area while keeping it visually subtle.

src/components/sessionHeader/sessionHeader.styles.scss
```diff
@@ -5,37 +5,35 @@ $iconSize: 24px;
...
+	min-height: 75px;
+	background-color: #fff;
 	box-sizing: border-box;
-	border-bottom: 1px solid rgba(80, 80, 80, 0.1);
-	margin-left: 1px;
-	// Ensure crisp 1px border rendering
-	-webkit-backface-visibility: hidden;
-	backface-visibility: hidden;
-	transform: translateZ(0);
+	margin-left: 40px;
```
shazia-k: Issue:  The header applies fixed 40px left and right margins across all viewports, combined with a nowrap supervision pill and larger mobile title text. This likely causes overflow or layout compression on mobile.
Suggestion: Apply the 40px margins only on desktop breakpoints, or introduce mobile-specific overrides for the header and supervision button.

shazia-k (commented):
@ALI-RUBASS there are some comments mostly related to css

shazia-k (approved):
Approved!

src/components/app/Routing.tsx
```diff
@@ -30,6 +31,7 @@ interface RoutingProps {
 export const Routing = (props: RoutingProps) => {
 	const location = useLocation();
 	const settings = useAppConfig();
+	const { untilL } = useResponsive();
```
shazia-k, line 34: untilL is destructured from useResponsive() but no longer used after the embedded-notifications change. Suggestion: Remove untilL and drop useResponsive here if nothing else needs it, to avoid lint noise and confusion.

src/components/app/Routing.tsx
```diff
@@ -80,7 +83,7 @@ export const Routing = (props: RoutingProps) => {
 					<NonPlainRoutesWrapper logoutHandler={() => props.logout()}>
 						<div
 							className={`app__wrapper ${
-								isEmbeddedNotificationsView
+								useEmbeddedNotificationsLayout
```
shazia-k, line 86: useEmbeddedNotificationsLayout is now isEmbeddedNotificationsView only (no !untilL). 
Issue: Behavior change: embedded layout applies at all breakpoints when the query flag is set. Suggestion: Confirm with product/QA (iframe / embedded notifications on desktop) that this is intended.

src/components/sessionsList/SessionsListToolbar.tsx
```diff
@@ -0,0 +1,552 @@
...
+								) : (
+									<div className="sessionsListToolbar__searchEmpty">
+										{tr(
+											'sessionList.toolbar.search.emptyPeople',
+											'No matching people found.'
+										)}
+									</div>
+								)
+							) : (
+								<div className="sessionsListToolbar__searchEmpty">
```
shazia-k, line 527: Non-people tabs always show “No results for this tab yet.” 
Issue: Tabs look fully interactive but are placeholders. 
Suggestion: Hide/disable unfinished tabs or implement minimal behavior so users are not misled.

shazia-k (commented):
@ALI-RUBASS there are some comments added

src/components/message/MessageItemComponent.tsx
```diff
@@ -67,7 +67,80 @@ import { VideoChatDetails, VideoChatDetailsAlias } from './VideoChatDetails';
...
+import { ReactComponent as ShieldIcon } from '../../resources/img/icons/shield.svg';
+import { ReactComponent as EyeIcon } from '../../resources/img/icons/eye.svg';
+import { ReactComponent as ArrowLeftIcon } from '../../resources/img/icons/arrow-left.svg';
+import { ReactComponent as StackVerticalCircleIcon } from '../../resources/img/icons/stack-vertical-circle.svg';
+import { ReactComponent as PenIcon } from '../../resources/img/icons/pen.svg';
+import { ReactComponent as ArrowForwardIcon } from '../../resources/img/icons/arrow-forward.svg';
+
+const ActiveKebabIcon = () => (
+	<svg width="28" height="32" viewBox="0 0 28 32" fill="none" aria-hidden>
+		<g clipPath="url(#active-kebab-clip)">
```
shazia-k, line 84: Duplicate SVG clipPath id (active-kebab-clip) on every kebab icon instance → invalid HTML and wrong clip reference. 
Suggestion: Use React.useId() (or similar) for unique id / url(#id) per instance.

src/components/message/MessageItemComponent.tsx
```diff
@@ -216,6 +300,24 @@ export const MessageItemComponent = ({
 		setIsExpanded(false);
 	}, [_id]);
 
+	useEffect(() => {
```
shazia-k, line 398: Outside-click closes the menu on mousedown only. 
Suggestion: Consider focusout / focus containment or document that behavior is intentional; add Escape.

src/components/message/MessageItemComponent.tsx
```diff
@@ -456,6 +606,90 @@ export const MessageItemComponent = ({
...
+			{
+				key: 'delete',
+				label: translate('message.menu.delete', 'Delete Message'),
+				icon: <MenuDeleteIcon />
+			}
+		],
+		[translate]
+	);
+
+	const handleActionMenuItemClick = useCallback(
```
shazia-k, line 1023: handleActionMenuItemClick only implements reply-thread; other keys close the menu with no action. Suggestion: Implement handlers, disable items, or remove/hide until ready.

src/components/message/MessageDisplayName.tsx
```diff
@@ -23,28 +23,38 @@ export const MessageDisplayName = ({
 }: MessageDisplayNameProps) => {
 	const { t: translate } = useTranslation();
 	const { activeSession } = useContext(ActiveSessionContext);
-
-	const subscriberIsModerator = isUserModerator({
+	// keep reference to session context for future role-based label behavior
+	isUserModerator({
```
shazia-k: isUserModerator(...) called with no use of the return value; comment says “future” behavior. 
Suggestion: Remove the call or assign/use the result; avoid no-op calls.

src/components/session/SessionStream.tsx
```diff
@@ -117,23 +117,27 @@ export const SessionStream = ({
...
-			
+
 			// console.log('🚀 MATRIX: Fetching messages from Matrix API:', matrixUrl);
-			
+
 			// Use raw fetch to see actual response
 			const accessToken = getValueFromCookie('keycloak');
-			const csrfToken = document.cookie.split('; ').find(row => row.startsWith('CSRF-TOKEN='))?.split('=')[1];
-			
+			const csrfToken = document.cookie
```
shazia-k, line 223: CSRF cookie value is parsed with split('=')[1] after split('; '). Any = inside the token (common for Base64 padding or encoded values) is truncated, so X-CSRF-TOKEN and X-WHITELIST-HEADER can be wrong → failed or flaky Matrix message loads (401/403) or CSRF rejection.

Suggestion:
parsed with split('=')[1] after split('; '). Any = inside the token (common for Base64 padding or encoded values) is truncated, so X-CSRF-TOKEN and X-WHITELIST-HEADER can be wrong → failed or flaky Matrix message loads (401/403) or CSRF rejection. | Parse the cookie by taking the substring after the first =, e.g. row.slice('CSRF-TOKEN='.length) or split with limit / regex, then decodeURIComponent if the server encodes it. Prefer a shared getCookie('CSRF-TOKEN') helper used everywhere.

src/components/session/SessionStream.tsx
```diff
@@ -143,87 +147,139 @@ export const SessionStream = ({
...
+			})
+				.then(async (response) => {
+					// console.log('🚀 MATRIX: Response status:', response.status);
+
+					const responseText = await response.text();
+					// console.log('🚀 MATRIX: Response body:', responseText);
+
+					if (response.status === 200) {
+						const responseData = responseText
+							? JSON.parse(responseText)
```
shazia-k, line 249: JSON.parse(responseText) on 200 responses with no local try/catch. Malformed JSON rejects the promise; outer .catch clears messages (line ~1951+) so users can see an empty thread with little signal.
Suggestion:
Medium (not as severe as cookie parsing): wrap JSON.parse in try/catch, log, and surface a retry / error state instead of silently null messages.

src/components/messageSubmitInterface/messageSubmitInterfaceComponent.tsx
```diff
@@ -1897,13 +2369,102 @@ const AudiencePersonIcon = () => (
...
+				}
+				const mapped = (Array.isArray(supervisors) ? supervisors : [])
+					.map((entry) => ({
+						id: `${entry?.supervisorConsultantId || ''}`.trim(),
+						username: `${entry?.supervisorUsername || ''}`.trim()
+					}))
+					.filter((entry) => entry.id || entry.username);
+				setSessionSupervisors(mapped);
+			})
+			.catch(() => {
```
shazia-k, line 2458: If apiGetSessionSupervisors fails, supervisors stay empty → restrictToSupervisionAudience is false → audience widens to full room + normal candidates (fail-open).
Suggestion:
Model load error explicitly: retry/error UI, or fail-closed for supervision (do not silently expand recipients). Adjust catch + gating so a failed load cannot unlock a broader audience than a successful one.

src/components/messageSubmitInterface/messageSubmitInterfaceComponent.tsx
```diff
@@ -2503,27 +3642,356 @@ const AudiencePersonIcon = () => (
...
+													fill="none"
+													xmlns="http://www.w3.org/2000/svg"
+												>
+													<rect
+														width="32"
+														height="32"
+														rx="16"
+														fill="#CC1E1C"
+													/>
+													<path
```
shazia-k, line 3880: SVG uses HTML-style attributes; React expects DOM property names on SVG.
Suggestions:
Use strokeWidth, strokeLinecap, strokeLinejoin (camelCase).

src/components/messageSubmitInterface/messageSubmitInterfaceComponent.tsx
```diff
@@ -260,205 +327,563 @@ export const MessageSubmitInterfaceComponent = ({
...
+			aria-hidden="true"
+		>
+			<path
+				d="M5.5 2.56667L1.28333 6.78333L0 5.5L5.5 0L11 5.5L9.71667 6.78333L5.5 2.56667Z"
+				fill="currentColor"
+			/>
+		</svg>
+	);
+
+	const AudienceSingleUserIcon = () => (
```
shazia-k, line 836: Issue: audience role SVGs redefined here as AudienceSingleUserIcon / AudienceSingleConsultantIcon / AudienceModeratorIcon.
Suggestion:
Move to shared components (e.g. components/icons/audience/…) or shared SVG assets and import in both places.

src/components/session/SessionItemComponent.tsx
```diff
@@ -1672,11 +1920,30 @@ export const SessionItemComponent = (props: SessionItemProps) => {
 			scrollToEnd(0, true);
 		}
 	};
+	const handleMobileNavigateBackClick = () => {
+		mobileListView();
+	};
+	const handleMobileNavigateStepDownClick = () => {
```
shazia-k, line 2599: This function scrolls by 60% of viewport height
Consider matching the same step as any existing “page down” behavior elsewhere, or make the factor a named constant with a short comment.

shazia-k (commented):
@ALI-RUBASS Comments added

src/components/sessionsList/SessionsList.tsx
```diff
@@ -61,10 +63,145 @@ import { useWatcher } from '../../hooks/useWatcher';
...
+}
+
+/**
+ * Anonymous-chat sessions are identified by the ORISO convention that the
+ * asker's keycloak username is prefixed `Anonymous-` (capital A — the
+ * pseudonym-registration flow writes that). Using just the prefix keeps the
+ * split clean even though all sessions in this deployment share the same
+ * registrationType + `postcode='00000'` on the DB side.
+ */
+function isAnonymousAskerSession(
```
shazia-k, line 127: Issue: isAnonymousAskerSession includes consultant?.username in the same candidate list as asker fields. If that string ever matched the Anonymous- convention, sessions could be filtered into the wrong chip (Chats vs Live Chat).
Suggestion: Restrict candidates to asker-only fields (e.g. user.username, session.askerUserName); remove consultant?.username unless product explicitly requires it.

src/components/pseudonym/AnonymousConsentGate.tsx
```diff
@@ -0,0 +1,159 @@
...
+export const AnonymousConsentGate: React.FC<AnonymousConsentGateProps> = ({
+	consentLabelHtml,
+	onAccept,
+	busy = false
+}) => {
+	const { t } = useTranslation();
+	const [rejected, setRejected] = useState(false);
+
+	return (
+		<div className="anonymousConsentGate" role="dialog" aria-modal="true">
```
shazia-k, line 85: Issue: Marked as role="dialog" + aria-modal="true" without focus trap, initial focus, or Escape handling—keyboard and AT users may not experience this as a proper, operable consent step.
Suggestion:
Add focus management (e.g. focus heading or primary action on open) and Escape to close/dismiss if spec allows, or change to a non-modal pattern and adjust ARIA so it matches behavior.

src/resources/i18n/de@informal/common.json
```diff
@@ -365,6 +365,28 @@
 				"headline": "Oder vereinbaren Sie jetzt einen Termin"
 			}
 		},
+		"audience": {
```
shazia-k, line 368: New onboarding strings were added for German; other active locales may lack keys and fall back to code defaultValue or mixed language.
Add the same keys to en/common.json (and any other shipped locales) so non-German users get proper copy.

PR #200, opened 2026-06-18
FE-M08 centralize matrix room id detection into isMatrixRoomId util

Kadyan25:
## What
- Added \src/utils/isMatrixSession.ts\ — exports \isMatrixRoomId(id)\ which checks for the \!\ prefix that identifies Matrix room IDs
- Replaced scattered inline \startsWith('!')\ checks across 9 components with imports of the utility: \IncomingVideoCall\, \messageSubmitInterfaceComponent\, \NotificationsCenter\, \SessionItemComponent\, \SessionStream\, \SessionHeaderComponent\, \SessionsList\, \SessionListItemComponent\, \useE2EE\
- Removed duplicate local implementations of the same check

## Why
Matrix room ID detection was duplicated across 9+ components with no shared definition. A single wrong check could silently break E2EE, message routing, or video calls for Matrix sessions. Centralizing into one util makes future changes to the detection logic a single-file edit. Part of audit finding FE-M08.

## Test
- [ ] Open a Matrix-backed counseling session — messages send and receive correctly
- [ ] E2EE indicator shows correctly for Matrix sessions
- [ ] Rocket.Chat-backed sessions (legacy) still load without regression


Kadyan25:
Superseded by #194 (merged) — same FE-M08 fix. PR #194 covers more files and the matrixRoomUtils.ts utility is already on dev. Closing to avoid duplicate utility.

PR #227, opened 2026-06-22
[codex] Fix public topics auth header

### LINKED ISSUE #226: Bug: login page can render blank when a stale auth cookie is present
## Summary

`https://app.oriso.org/login` can render as a blank white screen when the browser has a stale or invalid `keycloak` cookie. The page requests public topic data during login bootstrapping, but the frontend sends the stale bearer token with that public request. The API then returns `401 Unauthorized`, the unhandled `UNAUTHORIZED` rejection bubbles out of `fetchData`, and React does not render the login UI.

## Where it works

- Fresh/isolated browser session with no stale auth cookie.
- Incognito/private session.
- Direct public API request without an `Authorization` header:

```bash
curl -sS -o /tmp/oriso-topics-noauth.json -w '%{http_code}\n' https://api.oriso.org/service/topic/public/
# 200
```

## Where it fails

- Existing browser profile on `https://app.oriso.org/login` when a stale/invalid `keycloak` cookie is present.
- The login screen stays blank/white.
- Browser console shows:

```text
GET https://api.oriso.org/service/topic/public/ net::ERR_ABORTED 401 (Unauthorized)
Uncaught (in promise) Error: UNAUTHORIZED
    at fetchData.ts:269
```

## Reproduction evidence

A stale bearer token is enough to reproduce the API behavior:

```bash
curl -sS -o /tmp/oriso-topics-staleauth.json -w '%{http_code}\n' \
  -H 'Authorization: Bearer stale-token' \
  https://api.oriso.org/service/topic/public/
# 401
```

The same endpoint returns `200` without the auth header, so the public login bootstrap should not attach auth to this call.

## Expected behavior

Public login/bootstrap data should load without using stale user credentials. The login page should render even when old auth cookies are present.

## Actual behavior

The public topics request includes the stale bearer token, receives `401`, rejects as `UNAUTHORIZED`, and the login route remains a white screen.

## Workaround

Until the fix is deployed, clearing site data/cookies for `app.oriso.org` and `api.oriso.org` or opening the page in an incognito/private window avoids the stale bearer header.

Storypapst:
## Problem

Fixes #226.

The login page can render as a blank white screen when the browser has a stale or invalid `keycloak` cookie. During login bootstrapping, the frontend fetches public topic data from `/service/topic/public/`, but `fetchData` automatically attaches the stale bearer token unless the caller opts out. The API returns `401 Unauthorized` for that stale bearer token and the resulting `UNAUTHORIZED` rejection prevents the login UI from rendering.

## Solution

- Mark `apiGetTopicsData` as `skipAuth: true`, matching the other public bootstrap calls.
- Return the non-`EMPTY` rejection from the catch handler so real loading errors are not swallowed.

## Validation

- Confirmed live API behavior:
  - `GET https://api.oriso.org/service/topic/public/` without auth returns `200`.
  - The same request with `Authorization: Bearer stale-token` returns `401`.
- Ran a targeted `ts-node` regression harness; it fails before the fix because `skipAuth` is unset and passes after the fix.
- `npx eslint src/api/apiGetTopicsData.ts`
- `npx tsc --noEmit --pretty false`

## Notes

Cypress was not used for the committed validation because this checkout's Cypress support bundle fails before spec execution with `process is not defined` while importing `getApiBaseUrl.ts`, and the extension config also excludes the existing login spec. The targeted harness covers the changed API wrapper behavior directly.

### DIFF
--- a/src/api/apiGetTopicsData.ts
+++ b/src/api/apiGetTopicsData.ts
@@ -9,11 +9,12 @@ export const apiGetTopicsData = async (): Promise<TopicsDataInterface[]> => {
 		url: url,
 		rcValidation: false,
 		responseHandling: [FETCH_ERRORS.EMPTY],
-		method: FETCH_METHODS.GET
+		method: FETCH_METHODS.GET,
+		skipAuth: true
 	}).catch((error) => {
 		if (error.message === FETCH_ERRORS.EMPTY) {
 			return [];
 		}
-		Promise.reject(error);
+		return Promise.reject(error);
 	});
 };


Storypapst:
Closing as superseded.

The login white-screen (#226) is already fixed on dev by #238 (stale-token self-heal). This PR's two changes were reviewed:

- `return Promise.reject(error)` — a real fix (non-EMPTY load errors were being swallowed). Salvaged into #252.
- `skipAuth: true` on the public topics call — intentionally not adopted. It collides with #238, whose self-heal relies on the public topics call surfacing a 401 to detect and clear the stale token; skipping auth would suppress that 401, leave the stale cookie in place, and break #238's regression test.

Superseded by #238 (white-screen) + #252 (error propagation).

PR #233, opened 2026-06-23
fix: add NODE_ENV guards to DevToolbar flags at consummption sites (FE-M10)

Shirloin:
## Summary

Fixes FE-M10: two DevToolbar localStorage flags affected production
behavior without any environment guard at their consumption sites.

## Problem

The DevToolbar UI is hidden in production by default (`showDevTools='0'`),
but the flags it manages are still readable from `localStorage` directly
via browser console — no UI needed. Two flags had no `NODE_ENV` check
at their actual read sites:

- `disable 2fa_duty` — setting to `'1'` in production console allowed
  consultants to bypass the client-side 2FA setup obligation overlay
- `attachement_encryption` — setting to `'0'` in production console
  disabled client-side attachment encryption on the legacy RocketChat path

Note: `e2ee_disabled` was also listed as a concern but already has an
explicit production guard in `e2eeSettings.ts` — not affected.

## Changes Made

Added `process.env.NODE_ENV !== 'production'` guards at all 4 consumption
sites:

- **TwoFactorNag.tsx** (2 sites): `disable 2fa_duty` forced to `'0'`
  in production — 2FA duty enforcement always active
- **TwoFactorAuth.tsx** (1 site): `disable 2fa_duty` check short-circuits
  to `false` in production — consultant cannot see disable button via flag
- **MessageAttachment.tsx** (1 site): `attachement_encryption` forced to
  `1` in production — decryption always attempted for `t=e2e` attachments
- **messageSubmitInterfaceComponent.tsx** (1 site): same — encryption
  always enabled on RocketChat upload path if `isE2eeEnabled`

Using `!== 'production'` rather than `=== 'development'` so guards
work correctly when `NODE_ENV` is undefined in local environments
without explicit `.env` config.

## Testing

- TypeScript: 0 errors
- Production behavior: both flags forced to secure defaults
- Non-production (dev/staging): flags still read from localStorage as before

### DIFF
--- a/src/components/message/MessageAttachment.tsx
+++ b/src/components/message/MessageAttachment.tsx
@@ -58,7 +58,9 @@ export const MessageAttachment = (props: MessageAttachmentProps) => {
 	);
 	const audioRef = useRef<HTMLAudioElement | null>(null);
 	const [isAudioPlaying, setIsAudioPlaying] = React.useState(false);
-	const [audioDurationSec, setAudioDurationSec] = React.useState<number | null>(null);
+	const [audioDurationSec, setAudioDurationSec] = React.useState<
+		number | null
+	>(null);
 	const [audioCurrentTimeSec, setAudioCurrentTimeSec] = React.useState(0);
 
 	const decryptFile = useCallback(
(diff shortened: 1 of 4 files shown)


softwaredevzestgeek (approved):
Approved PR, please merge your changes

Issue #245, opened 2026-06-23
Newly created agencies silently invisible in registration (hardcoded consultingType=1 vs modality filter; topic gating)


Storypapst:
## Summary

Agencies created under newer tenants are silently invisible in the public registration flow, even when the admin panel shows them as "visible in registration". No error is shown to the end user. Live-verified on production (`app.oriso.org` / `api.oriso.org`), 2026-06-23.

There are **two independent problems** that produce the same symptom and must not be conflated.

## Product clarification (must be reflected in any fix)

`consulting_type` is **NOT** deprecated. It encodes the **counselling/chat modality** (live/proximity chat, 1:1, internal chats, soon self-help-group chat, future video). The in-code comment *"as consulting types are not used anymore"* in `ORISO-Admin/src/api/consultingtype/getConsultingType4Tenant.ts` is **wrong and dangerous** and must be removed. Verified: `/service/consultingtypes/basic` carries `groupChat.isGroupChat`, `isVideoCallAllowed`, `isAnonymousConversationAllowed`, etc. per type.

## A — Reproduced today (data/UX)

Registration is topic-first; an agency only appears under the exact topic(s) it is tagged with, and **13 of 16 ACTIVE topics currently have zero online agencies but are still selectable** → guaranteed "Keine Online-Beratungsstelle gefunden" dead-ends.

| Topic | Frontend call | Result |
|---|---|---|
| U25 Suizidprävention (3) | `GET /service/agencies?postcode=10965&topicId=3&consultingType=1` | 200 → 5 agencies incl. "Schwangerschaftberatung Deutschland" ✅ |
| Schwangerschaft (12) | `GET /service/agencies?postcode=10965&topicId=12&consultingType=1` | 200 → `[]` → no agency ❌ |

The reported agency `#248` is healthy (`offline:false, consultingType:1, topicIds:[3]`) and appears under its tagged topic (U25); it is absent under any other topic. That is the "only older agencies appear" perception.

## B — Latent systemic defect (highest risk)

The frontend hardcodes `consultingType: consultingType?.id || 1` (`src/containers/registration/hooks/useAgenciesForRegistration.ts:138`); the backend enforces `a.consulting_type = :type` (`:type` never null). Because `consulting_type` = modality, **any agency with modality ≠ 1 is silently dropped**. Verified:

```
GET /service/agencies?postcode=10965&topicId=3&consultingType=1 -> 5 agencies
[...]

PR #246, opened 2026-06-23
Hide topics with no agency from the registration topic step

Storypapst:
## Problem

Registration is topic-first. The backend agency query `INNER JOIN`s `agency_topic`, so a topic with no linked agency always returns an empty list and the user hits "Keine Online-Beratungsstelle gefunden". Today **all** active topics are offered, but on the live system only 4 of 16 topics have any agency — so picking one of the other 12 (e.g. Schwangerschaft, Schulden, Eltern & Familie) is a guaranteed dead-end.

## Fix

Filter the registration topic-selection list to topics that have at least one agency assigned, using the existing `GET /service/agencies/topics` endpoint (`apiGetTenantAgenciesTopics`).

- Applies only when **browsing** (no deep-linked topic/agency/consultant) so deep links are never broken.
- **Fails open:** on request error or empty coverage, topics are left unchanged, so registration can never become unusable due to this filter.
- **Self-correcting:** as agencies get linked to topics, those topics automatically reappear.

Logic extracted into a pure helper `filterTopicsByAgencyCoverage` for clarity and testability.

## Validation

- `tsc --noEmit` clean (0 type errors).
- Behaviour confirmed against live coverage data: `GET /service/agencies/topics` currently returns topics [1, 2, 3, 7], so the 12 empty topics would be hidden until agencies are linked.

> Note: this repo's only test harness is Cypress e2e (no unit runner), so the helper is structured as a pure function for a future unit test / Cypress spec.

## Context

Part of "newly created agencies invisible in registration" — see OpenResilienceInitiative/ORISO-Frontend#245. Complementary backend fix (do not wipe agency-topic links on update): OpenResilienceInitiative/ORISO-AgencyService#39. Next step is to backfill the missing `agency_topic` links so the relevant topics regain coverage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Topic selection in the registration flow now filters topics based on agency coverage availability. Topics without available agency services are filtered out during the browse flow when no preselected values are active. The system gracefully falls back to showing all topics if coverage data is unavailable.

### DIFF
--- a/src/components/registration/topicSelection/TopicSelection.tsx
+++ b/src/components/registration/topicSelection/TopicSelection.tsx
@@ -28,6 +28,8 @@ import {
 } from '../../../globalState';
 import { apiGetTopicGroups } from '../../../api/apiGetTopicGroups';
 import { apiGetTopicsData } from '../../../api/apiGetTopicsData';
+import { apiGetTenantAgenciesTopics } from '../../../api/apiGetTenantAgenciesTopics';
+import { filterTopicsByAgencyCoverage } from './filterTopicsByAgencyCoverage';
 import {
 	TopicsDataInterface,
 	TopicGroup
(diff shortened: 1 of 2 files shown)


shazia-k (approved):
No unit tests for filterTopicsByAgencyCoverage (empty array, partial coverage, null input)
It would be great to implement test cases

PR #250, opened 2026-06-24
feat(timeline): route-derived list-item selection controller (WP-06 Slice 0b)

Storypapst:
## WP-06 Activity Timeline — Slice 0b (Global selection controller)

One **route-derived source of truth** for which list item is active, replacing the per-component flags that drift apart and let several items go "red" at once. Robustness fix, **not** a new design.

### What's in here
- **`src/utils/listItemSelection.ts`** — pure `deriveActiveSelection(pathname)` + `isListItemActive(selection, identity)`. Handles every session route (consultant `sessionView`/`sessionPreview`, asker `view`; RC `:rcGroupId/:sessionId`, Matrix `session/:sessionId`, enquiry `write/:sessionId`) and **normalises string vs number** so a numeric `item.id` matches the string route param.
- **`src/hooks/useActiveListItem.ts`** — thin `useLocation`-based hook exposing `{ selection, isActive }`.
- **`SessionListItemComponent`** (conversation list **and** request list) now derives `isChatActive` from the hook instead of `activeSession.rid === groupIdFromParam || activeSession.item.id === sessionIdFromParam`.
- **New behaviour**: entering a chat room now marks that conversation active — the old comparison matched the literal `session` route segment and never lit up Matrix rooms.
- **`src/resources/styles/_listItemSelection.scss`** — shared tokens + mixins (`list-item-active-treatment`, `list-item-focus-treatment`), imported globally via `settings.scss`.

### Visual treatment (normalise, don't restyle)
- **Conversation / request lists**: unchanged. Their existing active/focus values are now **sourced from the shared tokens** (same colours) and the focus ring uses the shared mixin — zero visual change.
- **Activity Timeline**: its one-off `#fff6f7` active row is normalised to the **canonical treatment** (lightened background + **2px accent ring**) and gains an independent `:focus-visible` ring, so it matches the conversation/request lists.
- Active = lightened bg + 2px accent ring + red action button; resting = 1px secondary; `focus-visible` is independent and layers on top.

### Invariant
Active identity is derived **once** from the URL; every item compares against it, so two items can't both be active (sessionId/groupId are unique per conversation). Covered by the unit spec, including the exactly-one-active count and the enter-room-activates case.

### Scope guardrails (explicitly **not** in this slice)
- No event-descriptor registry (that's **Slice 0a**, separate PR — disjoint file set, no overlap).
- No restyle of the conversation/request lists; no Handover/Calls/Drafts/backend work.

### Verification
- `npx tsc` — 0 errors; `eslint` on touched files — 0 errors (only pre-existing warnings).
- `stylelint` — my additions are clean (remaining errors pre-exist on `dev`).
- `CI=false npm run build` — succeeds.
- Selection logic verified green: route derivation, normalisation, enter-room-activates, and exactly-one-active all pass.

Refs: WP-06 Activity Timeline; CONTEXT.md "Active item".

### DIFF
--- a/cypress/e2e/listItemSelection.cy.ts
+++ b/cypress/e2e/listItemSelection.cy.ts
@@ -1,0 +1,123 @@
+/**
+ * WP-06 Activity Timeline — unit coverage for the route-derived list-item
+ * selection (Slice 0b). Pure-logic spec: imports the selection utility directly
+ * and asserts with chai's `expect`. No app visit / backend required.
+ */
+
+import {
+	deriveActiveSelection,
+	isListItemActive,
+	ListItemIdentity
+} from '../../src/utils/listItemSelection';
+
+describe('WP-06 list-item selection controller', () => {
+	describe('deriveActiveSelection', () => {
+		it('has no active item on bare list views or the timeline', () => {
+			['/sessions/consultant/sessionView', '/sessions/user/view', '/notifications', '/', ''].forEach(
+				(p) => {
+					expect(deriveActiveSelection(p), p).to.deep.equal({
+						groupId: null,
+						sessionId: null
+					});
+				}
+			);
+		});
+
+		it('derives the sessionId from a Matrix session route (groupId null)', () => {
+			expect(
+				deriveActiveSelection('/sessions/consultant/sessionView/session/123')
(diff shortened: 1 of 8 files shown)


Storypapst:
Tracked in #251

PR #249, opened 2026-06-23
feat(timeline): event-descriptor registry (WP-06 Slice 0a)

Storypapst:
## WP-06 Activity Timeline — Slice 0a (Event-descriptor registry)

Additive foundation for the Activity Timeline. Introduces a **single registry** that turns a typed, text-free server record into a rendered timeline card, and wires it into `NotificationsCenter` to replace the generic bell + system/message tag with per-type icons and family labels.

### What's in here
- **`src/components/notificationsCenter/eventDescriptors/`** — pure registry mapping `event_type → { icon, family, category, i18n title/text template, resolveActionTarget(params) }`.
  - `types.ts`, `registry.ts`, `renderEventStrings.ts` are pure (no React/SVG imports) and unit-testable in isolation; `icons.tsx` is the only file importing SVGs. `tsc` enforces icon-map completeness (`Record<EventIconId, …>`).
  - **Seeds all listed types**: the 6 emitted today (`inquiry.accepted`, `message.new`, `thread.reply.new`, `supervisor.added/removed/assigned`, `counselor.renamed`) **plus** descriptors for the designed families **Requests, Messages, Drafts, Handover, Calls, System** (Appointments deferred).
  - **Action-target rule = origin** (per `CONTEXT.md`): chat → conversation; request-origin → request view; draft → resume via `forcedScopeKey`; live call → **Join**. Targets are **pure data** — no navigation or call-stack wiring (deferred to Slices 4/5).
- **`NotificationsCenter.tsx`** — now renders each card's icon/family/title/text from the descriptor. Every visible string comes from i18n templates client-side (**ADR-AT-01**: the server record is text-free). While the backend still sends `title`/`text` (pre-Slice-2), those are used only as the i18n `defaultValue`, so nothing regresses and **no server-side rendered text is introduced**.
- **i18n** — `notifications.events.*` + `notifications.families.*` added to `en` and `de`; other locales fall back to `de` (`FALLBACK_LNG`).
- **Tests** — `cypress/e2e/eventDescriptors.cy.ts`: a pure-logic spec covering every seeded type (family/category/icon/templates), the origin action-target kinds, fallback for unknown types, i18n key presence (en+de), and client-side rendering. Runs under the repo's Cypress harness.
- **`CONTEXT.md`** — the Activity Timeline glossary referenced by the registry's code comments and the ADRs.

### Scope guardrails (explicitly **not** in this slice)
- No global selection controller (that's **Slice 0b**, separate PR — this PR leaves `--active` untouched).
[...]

PR #253, opened 2026-06-24
test(timeline): vitest unit coverage + wire timeline to selection controller (WP-06 Slice 0 follow-up)

Storypapst:
## What & why

Follow-up that closes two **Definition-of-Done** gaps left by the already-merged WP-06 Slice 0 PRs (#249 registry / #250 selection controller). The registry and selection controller themselves are solid and unchanged; this PR only addresses where they fell short of the slice's DoD.

### Gap 1 — CI-enforced unit coverage
Both Slice 0 PRs shipped their pure-logic tests as `cypress/e2e/*.cy.ts`. CI runs **vitest** (`npm run test:unit`, scoped to `src/**/*.test.{ts,tsx}`) — it never picks up `cypress/e2e/`, and those specs can't run bare either. So the registry + selection logic shipped with **zero CI-enforced unit coverage**, despite specs existing.

- Port both specs to **vitest** next to their source:
  - `src/utils/listItemSelection.test.ts`
  - `src/components/notificationsCenter/eventDescriptors/registry.test.ts`
- Delete the now-dead `cypress/e2e/{eventDescriptors,listItemSelection}.cy.ts` duplicates.
- Coverage **increases**: adds a `17 seeded types` assertion, a null/undefined-selection case, and a full suite for the new `pickActiveItemKey` helper.

### Gap 2 — wire the Activity Timeline to the shared selection controller
Slice 0b wired the conversation/request list (`SessionListItemComponent` → `useActiveListItem`) but the timeline still chose its active card with a bespoke `selectedNotificationId === item.id` flag — exactly the per-component duplication the slice set out to remove.

- Add a pure, reusable helper `pickActiveItemKey` to `utils/listItemSelection.ts`. It returns **exactly one** active key by construction: the first feed item whose origin conversation matches the route-derived selection wins; otherwise the in-page master-detail selection is used.
- `NotificationsCenter` now derives a single `activeNotificationId` through it and uses that for both the `--active` highlight and the detail-pane lookup.

`NotificationsCenter` is mounted only at `/notifications` (`exact: true`), where the route selection is always empty — so `activeNotificationId === selectedNotificationId` on every reachable render and there is **zero behavioral/visual change**. The change only removes the duplicated flag and unifies the source of truth, so the timeline is ready for (and cannot disagree with) the conversation/request lists. **No SCSS changed** — the active treatment already uses the shared `_listItemSelection` mixins from #250.

## DoD mapping
- ✅ Registry + selection logic now have **CI-enforced unit coverage** (vitest).
- ✅ Selection controller **wired into the timeline** (in addition to the conversation + request lists already covered by #250's shared `SessionListItemComponent`).
- ✅ Exactly-one-active invariant demonstrated by unit tests (incl. "entering a chat room activates that conversation" and the `pickActiveItemKey` exactly-one cases).
- ✅ No visual restyle.

## Verification
- `npm run test:unit` → **113 passing** (7 files; +34 from this PR).
- `tsc` → **0 errors**.
- `eslint src/...` → **0 errors** (3 warnings are pre-existing and unrelated: dead `getChatKey`, two hook-dep warnings on untouched code).
- `CI=false npm run build` → **OK**.
- Independent 3-lens adversarial review (React/behavioral regression · exactly-one invariant · test fidelity) → all **ship**.

## Conscious tradeoff
The deleted cypress registry spec asserted `EVENT_ICONS[d.icon]` resolves at runtime. The vitest port validates each descriptor's `icon` against the declared `EventIconId` set instead of importing `icons.tsx`, because vitest has no SVG (`ReactComponent`) transform here and a stub would be trivially always-true. The real invariant — every descriptor icon id has an `EVENT_ICONS` entry — is already enforced by the type system (`EVENT_ICONS: Record<EventIconId, SvgIcon>`, tsc gated in CI), and `getEventIcon` falls back to the system icon. Adding SVG-transform infra to vitest was deliberately left out of scope.


Storypapst:
Rebased onto latest `dev` (force-pushed `fa3a7f1` → `b936fb4`) to resolve a conflict introduced by #247, which refactored `NotificationsCenter` (removed the inline chat-preview/reply feature, the dead `getChatKey` helper, and `resolveRoomRef`).

Adapted the timeline wiring to the new component with **no change in approach**:
- The active-card identity extractor now uses `sessionId` only (`resolveRoomRef` no longer exists upstream). This also removes the dormant "path-segment as pseudo-groupId" smell flagged in pre-merge review — strictly cleaner.
- Dropped the now-obsolete `chatReplyText` reset edit (that state was removed by #247).

Net change vs `dev` is unchanged in spirit: import the shared controller, resolve a single `activeNotificationId` via `pickActiveItemKey`, and use it for both the `--active` highlight and the detail-pane lookup. Still zero behavioral/visual change on the standalone `/notifications` route, still no SCSS touched.

Re-verified on the rebased commit: **tsc 0 errors · eslint 0 problems** (the previously-noted pre-existing warnings are gone — #247 removed that code) **· vitest 116 passing · production build OK.**

PR #279, opened 2026-06-25
Fix redirected enquiry composer submit

Storypapst:
## Problem
After registration, askers are redirected to the normal session view. The composer can then receive `MY_SESSION` as the list type even though the session is still in enquiry status, so the send action does not route through the enquiry submission path.

## Solution
Add a shared `isAskerEnquirySubmission` predicate for enquiry routing and upload gating. It treats either the enquiry list type or `STATUS_ENQUIRY` as an asker enquiry submission, while keeping anonymous live chat excluded.

## Validation
- `npm run test:unit -- src/components/messageSubmitInterface/messageEncryptionMode.test.ts` - 8 tests passed
- `npm run lint:scripts` - 0 errors, 124 existing warnings
- `npm run test:unit` - 140 tests passed

### DIFF
--- a/src/components/messageSubmitInterface/messageEncryptionMode.test.ts
+++ b/src/components/messageSubmitInterface/messageEncryptionMode.test.ts
@@ -1,8 +1,10 @@
 import { describe, expect, it } from 'vitest';
 import {
+	isAskerEnquirySubmission,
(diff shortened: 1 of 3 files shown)


Storypapst:
Tracked in #280

Issue #285, opened 2026-06-26
Composer send button does not submit Matrix replies


Storypapst:
## Background
Pre-Dev Matrix enquiry sessions can be opened by a counselor, but clicking the composer send control leaves the typed reply in the editor and only updates drafts. No Matrix `/send/m.room.message` request is emitted.

## Expected behavior
Clicking the send control in the session composer submits the reply once through the existing send pipeline.

## Actual behavior
The send control is a native button, but the surrounding form has no submit handler. In the Pre-Dev E2E session, click and keyboard activation left the reply as a draft and did not create a Matrix send request.

## Evidence
- Pre-Dev session `103024` showed the typed counselor reply still in the composer after send attempts.
- UserService logs only showed draft PATCH/GET calls for the session scope.
- Synapse logs showed no `/send/m.room.message` request for the room during the attempts.

## Fix direction
Route the real composer send control through native form submission with `preventDefault()`, while keeping the shared send button default behavior for non-form contexts.

## Validation
- `npx vitest run src/components/messageSubmitInterface/SendMessageButton.test.tsx --passWithNoTests`
- `npm run test:unit`
- `npm run build`

Storypapst:
PR: #286

Storypapst:
Additional PR: #287

Storypapst:
Additional PR: #288

Storypapst:
Additional PR: #289

Storypapst:
Pre-Dev E2E validation passed after #291.

Validated on `2026-06-26` against deployed frontend digest `ghcr.io/openresilienceinitiative/oriso-frontend@sha256:8e967d537001efaf0673c06ecfed8ad8598cf2a47c5e795dbbc1662f9cc33694`.

Evidence:
- Fresh asker: `codexasker20260626020403`
- Session: `103025`
- Matrix room: `!aZWDxtSzBzaSPRkizR:91.99.183.160`
- Agency matching for topic `9`, postcode `12345`, consulting type `1` returned agency `263`.
- First enquiry message returned `201`.
- Consultant accepted the enquiry via UserService.
- Consultant UI loaded bundle `/static/js/app.da0b1503.js`.
- Consultant sent a reply from the frontend composer; network captured `PUT /_matrix/client/v3/rooms/!aZWDxtSzBzaSPRkizR%3A91.99.183.160/send/m.room.message/...` with `200`.
- Composer cleared after send.
[...]

PR #287, opened 2026-06-26
fix: persist refreshed matrix token

Storypapst:
## Problem
Pre-Dev Matrix counselor replies still stayed in the composer after the submit fix. Browser instrumentation showed the native submit path fired and React handled it, but the send path only saved drafts.

The Matrix token metadata in localStorage stayed expired after refresh. Each send attempt refreshed the Matrix client again, replacing the synced client before the room encryption state was available.

## Solution
- Persist refreshed Matrix access token and Matrix user id in localStorage.
- Persist `matrix_token_expires_at` from the token response `expiresInMs` value.
- Add unit coverage for the Matrix login-data persistence contract.

Relates to #285

## Validation
- `npx vitest run src/components/sessionCookie/getMatrixAccessToken.test.ts --passWithNoTests`
- `npm run test:unit`
- `npm run build`

## Remaining risk
Pre-Dev E2E verification is still required after merge and deploy.

### DIFF
--- a/src/components/sessionCookie/getMatrixAccessToken.test.ts
+++ b/src/components/sessionCookie/getMatrixAccessToken.test.ts
@@ -1,0 +1,55 @@
+// @vitest-environment jsdom
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { persistMatrixLoginData } from './getMatrixAccessToken';
+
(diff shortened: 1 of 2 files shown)


Storypapst:
Tracked in #285

PR #288, opened 2026-06-26
fix: send matrix messages through context client

Storypapst:
## Problem
Matrix counselor replies still did not leave the composer after the submit and token persistence fixes. Runtime probes showed the form submit and React handler were reached, but no Matrix `/send/m.room.message` request was emitted.

`MessageSubmitInterfaceComponent` already has the live `matrixClientService` from React context, but `apiSendMessage` only looked up a module-level registry singleton. In lazy chunk/runtime paths this can diverge from the synced context client.

## Solution
- Allow `apiSendMessage` to receive an explicit Matrix client service.
- Pass the context `matrixClientService` from the composer into the Matrix send helper.
- Keep the existing registry fallback for other callers.
- Add unit coverage proving Matrix sends use the provided service.

Relates to #285

## Validation
- `npx vitest run src/api/apiSendMessage.test.ts --passWithNoTests`
- `npm run test:unit`
- `npm run build`

## Remaining risk
Pre-Dev E2E verification is still required after merge and deploy.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Matrix message sending so messages are routed through the active client service more reliably.
  * Local echo sending now stays in sync with the current message service, reducing the chance of failed sends when the service changes.

* **Tests**
  * Added coverage for Matrix-backed message sending to verify successful sends return the expected event ID.

### DIFF
--- a/src/api/apiSendMessage.test.ts
+++ b/src/api/apiSendMessage.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it, vi } from 'vitest';
 import { apiSendMessage } from './apiSendMessage';
+import { getMatrixClientService } from '../services/matrixClientRegistry';
 
 vi.mock('../resources/scripts/endpoints', () => ({
 	endpoints: {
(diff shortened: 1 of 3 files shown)


Storypapst:
Tracked in #285

PR #320, opened 2026-06-26
fix(registration): skip welcome screen

Storypapst:
## Problem

The public registration welcome page is no longer aligned with the current registration flow. Users who choose registration should land directly on topic selection instead of an intermediate explanation screen.

## Solution

- Redirect unknown/welcome registration entries to the first active registration step.
- Keep the back/next URL calculation inside the available registration steps instead of falling back to `welcome`.
- Move the two remaining advice benefits into the post-registration loader, where they bridge the real app bootstrap before the enquiry view opens.

## Validation

- `npm run test:unit -- src/components/app/registrationLoader/RegistrationLoader.test.tsx`
- `npm run test:unit`
- `npm run lint:scripts` (existing warnings only)
- Production build with PreDev-like env

## Notes

Local rendered QA was blocked by the CRA dev proxy TLS chain and then by the in-app browser URL policy after a localhost failure. Final rendered validation must be performed on PreDev after the image built from `dev` is deployed.

### DIFF
--- a/src/components/app/registrationLoader/RegistrationLoader.test.tsx
+++ b/src/components/app/registrationLoader/RegistrationLoader.test.tsx
@@ -36,6 +36,12 @@ describe('RegistrationLoader', () => {
 
 		expect(screen.getAllByText('registration.loader.copy')).toHaveLength(2);
 		expect(screen.queryByText('registration.loader.ready')).toBeNull();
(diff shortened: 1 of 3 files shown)


Storypapst:
Tracked in #321

PR #361, opened 2026-07-03
Add public DPA signature route

Storypapst:
## Problem
External DPA signers need a public route that can confirm a single-use DPA token without requiring a platform account.

## Solution
- Add `/dpa-sign/:token` before the authenticated app catch-all.
- Add a public DPA signature API helper.
- Render required signer fields and acceptance checkbox.
- Show a stable error for expired or already-used tokens.

## Validation
- `npm run test:unit -- src/api/apiDpaSignature.test.ts`
- `./node_modules/.bin/tsc --noEmit --pretty false`
- `npm run build`
- Local browser smoke against the built app: valid token success and expired token error.

## Coordination
Depends on the TenantService DPA signature fields PR. Merge before deploying Frontend dev image to Pre-Dev.

### DIFF
--- a/src/api/apiDpaSignature.test.ts
+++ b/src/api/apiDpaSignature.test.ts
@@ -1,0 +1,71 @@
+import { describe, expect, it, vi } from 'vitest';
+import { apiConfirmDpaSignature, DPA_SIGN_ERRORS } from './apiDpaSignature';
+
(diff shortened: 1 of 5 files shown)


Storypapst:
Tracked in OpenResilienceInitiative/ORISO-Frontend#362.

PR #364, opened 2026-07-03
feat: user-facing department legal texts + configurable registration consulting type

Storypapst:
## What users get

- **Agencies of other modalities can show up in registration again.** The agency search no longer has the consulting type hard-coded to 1 (FE#245). A selected consulting type always wins (even id 0, which the old code silently replaced with 1), and deployments can override the fallback via the new app config setting `registration.defaultConsultingTypeId`. The default stays 1, so nothing changes for existing tenants unless they opt in (the backend requires the parameter, so it cannot be dropped entirely).
- **Departments show their own privacy and legal texts.** When the selected agency has a department for the selected topic with published texts, the agency details panel shows the department (topic name) and an expandable "Datenschutzhinweise der Beratungsstelle" section. It loads the text only when the user opens it.
- **The consent step prefers the department text.** On the account step, when the department has a published data privacy policy, that text is shown. If it cannot be loaded, the tenant text is used instead.
- **Translated legal texts are marked.** Machine-translated texts show a note: "Maschinell übersetzt — rechtlich verbindlich ist die deutsche Fassung", with a link to view the original. If a text is missing in the user's language, the original language is shown (with a short note) instead of showing nothing. This applies to the `/impressum` and `/datenschutz` pages and to department texts.

## Backend counterpart

AgencyService PR OpenResilienceInitiative/ORISO-AgencyService#90:

- `departments[]` (`topicId`, `hasPublishedDpp`, `hasPublishedImprint`) on the agency search response
- `GET /agencies/{agencyId}/topics/{topicId}/legal` with published content only (JSON language→HTML map, `<lang>__meta` keys mark machine translations)

**The endpoint is live only after #90 merges.** Until then the feature degrades to today's behavior:

- older backends never send `departments[]`, so no new UI renders at all
- if the flags exist but the legal endpoint answers 404 or fails, the API helper resolves to `null` (no error-page redirect) and the tenant text is used — covered by unit tests

## Changes

- `src/containers/registration/hooks/useAgenciesForRegistration.ts` — de-hardcoded consulting type + config fallback
- `src/utils/legalContent.ts` — pure `resolveLegalContent(map, uiLang)` + consent preference logic
- `src/components/legalContent/LegalContentRenderer.tsx` — reusable renderer with language/MT notices (wired into `/impressum`, `/datenschutz` via `LegalPageWrapper`)
- `src/components/departmentLegal/DepartmentLegalSection.tsx` — lazy-loading department legal section (agency details panel + consent step)
- `src/api/apiGetDepartmentLegal.ts` — public endpoint client with graceful 404 handling
- i18n keys for de, de@informal, en

## Tests

- 37 new vitest cases (resolution util, hook parameter, renderer notices/toggle, department section incl. 404 fallback, API graceful degradation)
- `tsc`, `npm run test:unit` (357 passed), production build: all green

### DIFF
--- a/src/api/apiGetDepartmentLegal.test.ts
+++ b/src/api/apiGetDepartmentLegal.test.ts
@@ -1,0 +1,71 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { apiGetDepartmentLegal } from './apiGetDepartmentLegal';
+import { fetchData, FETCH_ERRORS } from './fetchData';
+
+vi.mock('./fetchData', async () => {
+	const actual =
+		await vi.importActual<typeof import('./fetchData')>('./fetchData');
+	return {
+		...actual,
+		fetchData: vi.fn()
+	};
+});
+
+vi.mock('../resources/scripts/endpoints', () => ({
+	endpoints: {
+		agencyDepartmentLegal: (agencyId: number, topicId: number) =>
+			`https://api.test.local/service/agencies/${agencyId}/topics/${topicId}/legal`
+	}
+}));
+
+describe('apiGetDepartmentLegal', () => {
+	afterEach(() => {
+		vi.clearAllMocks();
+	});
+
+	it('resolves the published legal texts', async () => {
+		const legal = {
+			dpp: { content: '{"de":"<p>Hallo</p>"}' },
+			imprint: { content: null }
+		};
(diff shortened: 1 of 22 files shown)


Storypapst:
Closing: this PR's content landed on dev via #368, which was stacked on this branch and included all its commits (shazia-k's approval on #368 covered the combined diff). No code was lost — the diff here is now empty against dev.

PR #365, opened 2026-07-03
fix: preserve group chat session navigation

Storypapst:
## What changed
- Keep the consultant on the newly created group chat session route after creation.
- Refresh session rooms as a best-effort follow-up without blocking navigation when the refresh fails.
- Remove the success-overlay detour so group chat creation lands directly in the session view.

## Why
The rescued local fix prevents a successful group chat creation from looking broken when the post-create session refresh is slow or fails. The user gets routed to the created session instead of being left on the form.

## Validation
- `npm ci --legacy-peer-deps`
- `npm run lint:scripts` (0 errors, existing warnings only)


Storypapst:
Tracked in #366.

PR #385, opened 2026-07-08
feat(agents): introduce orchestration skills and documentation for task management. 

nikunjdecyb:
- Added new orchestration skills: `goal-loop`, `planner`, `verifier`, `security-auditor`, and `codebase-explorer` to streamline task execution and validation.
- Enhanced AGENTS.md with orchestration workflow details, including task triage, delegation, and approval gates.
- Created documentation for each skill to define their purpose and usage within the task workflow.
- Established a structured approach for task documentation in `docs/cursor-orchestrator/` to improve clarity and consistency.


.gitignore
```diff
@@ -52,4 +52,4 @@ graphify-out/*
 # chromatic build log
 build-storybook.log
 docs/*
```
shanzaeimran, line 54: `docs/*` here ignores everything under `docs/`, but this PR's whole orchestration
system writes its per-task trail into `docs/cursor-orchestrator/YYYY-MM-DD_.../`
(problem-brief, spike, progress-log, etc.). `index.md` only landed because it was
force-added; every future task artifact will be silently gitignored and can't be
committed without `-f`, which defeats the documentation-trail this feature exists
to produce.

Suggest adding a negation right after the `docs/*` line:

    docs/*
    !docs/cursor-orchestrator/

.cursor/skills/pr-prep/SKILL.md
```diff
@@ -0,0 +1,23 @@
...
+- **Summary**: what changed and why, 2–4 sentences
+- **Files changed**: from `git diff --stat origin/dev...HEAD`
+- **Test evidence**: link to `04-test-evidence.md`, plus the headline results inline
+- **Screenshots**: embed if UI changed
+- **Risks / follow-ups**: anything reviewers should watch, deferred work
+- **Security**: link `05-security-review.md` if one was produced
+
+PR conventions for this repo:
+
+- Base branch is `dev`; compare against `origin/dev`.
```
shanzaeimran: Same base-branch inconsistency as goal-loop — this says `dev`/`origin/dev` while
git-workflow.mdc and the PR target say `pre-dev`. Please reconcile to one base.

.cursor/skills/goal-loop/SKILL.md
```diff
@@ -0,0 +1,67 @@
...
+
+# Goal Loop
+
+Run the whole loop in one turn. Do not stop between iterations to ask "shall I continue?" — only stop at the exit conditions below.
+
+## Setup (once, before iteration 1)
+
+1. Normalize the goal with the `problem-intake` skill → `00-problem-brief.md`. The acceptance criteria in that brief are the loop's exit test. If the goal is one sentence, write criteria yourself from context; ask the user only for genuinely blocking unknowns.
+2. If non-trivial (more than one file, new behavior, UI, security, or unclear requirements): run the `planner` subagent → `01-spike.md` and `02-implementation-plan.md` (see `spike-doc` and `task-implementation-doc` skills).
+3. Git: check `git status`; confirm with user if the tree is dirty. Branch from `dev` as `cursor/<ticket-or-feature>/<short-slug>`. Never implement on `dev`.
```
shanzaeimran: Base-branch guidance contradicts the rest of the PR. Here (and in
pr-prep/SKILL.md) it says branch from / compare against `dev`, but
git-workflow.mdc says `pre-dev`, and this PR itself targets `pre-dev`. An agent
following goal-loop will branch off the wrong base. Please make all references
consistent (pre-dev, based on the PR target).

src/components/messageSubmitInterface/richtextHelpers.ts
```diff
@@ -105,8 +105,7 @@ export const markdownToDraftDefaultOptions = {
 	remarkablePreset: 'commonmark',
 	remarkableOptions: {
 		html: true,
-		breaks: true,
-		linkify: true
```
shanzaeimran: Removing `linkify` disables markdown auto-linkification for rendered messages
(this options object is consumed in MessageItemComponent rendering, where
`urlifyLinksInText` runs afterward). Can you confirm `urlifyLinksInText` covers
everything remarkable's linkify did — e.g. bare `www.` URLs and email-style
tokens — so we don't regress those into plain text?

src/components/message/MessageItemComponent.tsx
```diff
@@ -861,27 +861,32 @@ export const MessageItemComponent = ({
...
-						sanitizeHtmlDefaultOptions
-					)
-				: ''
-		);
+							),
+							sanitizeHtmlDefaultOptions
+						)
+					: ''
+			);
+		} catch {
```
shanzaeimran: Nit: the empty catch swallows the parse error with no trace. Given the fallback
silently renders raw markdown, a single `logger.debug(...)` in the catch would
keep this diagnosable without adding console noise.

shanzaeimran:
@nikunjdecyb 
Overall this looks solid — the matrix-js-sdk log-suppression module, its tests, and the `createClient` wiring are clean, and the console.log cleanup is welcome.

Two things I'd fix before merge, both in the new orchestration docs (they break the feature's own workflow rather than runtime behavior):

1. `.gitignore` still has `docs/*`, so the entire `docs/cursor-orchestrator/`  task trail this feature produces will be gitignored. Add `!docs/cursor-orchestrator/`.
2. Base-branch instructions contradict each other — goal-loop and pr-prep say `dev`, git-workflow says `pre-dev`, and the PR targets `pre-dev`. Pick one.

Non-blocking: confirm removing `linkify` doesn't regress bare-URL rendering, the empty catch in MessageItemComponent could log at debug level, and double-check the dev-only relative-origin logic against Storybook. Also consider splitting the orchestration tooling from the matrix/CORS/logging fixes in future PRs — they're unrelated and it'd make review/revert cleaner.

.gitignore
nikunjdecyb: @shanzaeimran - this was intententionaly made. because if we didn't make it then lot of task which is useless for other developer came and make our code heavy. so this cursor orchestoator and docs for local use only no need to push on the GitHub.

.cursor/skills/goal-loop/SKILL.md
nikunjdecyb: done

src/components/messageSubmitInterface/richtextHelpers.ts
nikunjdecyb: done

src/components/message/MessageItemComponent.tsx
nikunjdecyb: done

shanzaeimran (approved):

PR #389, opened 2026-07-08
Fix audio/video call failing with "Cannot send Matrix event before the target room is confirmed as an encrypted Matrix room"

shanzaeimran:
### Problem

Clicking the audio or video call button showed a browser alert:

> Failed to start call: Cannot send Matrix event before the target room is confirmed as an encrypted Matrix room

### Root cause

The Matrix E2EE teardown moved the platform to an app-level encryption layer:

- **ORISO-UserService** now creates all session rooms with an empty `initial_state` — deliberately **without** the `m.room.encryption` state event (`MatrixRoomClient.java`: *"Matrix E2EE is disabled because the frontend applies its own encryption layer"*).
- The frontend Matrix client accordingly never initializes crypto (no `initRustCrypto`).

However, the call paths still carried guards from the E2EE era: `assertMatrixRoomEncrypted()` required every target room to have an `m.room.encryption` state event before sending call events. Since session rooms no longer have that event, **every call start threw** — the alert message is that assert's error, surfaced by `CallManager.startCall()`'s catch handler.

Two frontend room-creation paths (`CallManager.createElementCallRoom`, `matrixClientService.createDirectMessageRoom`) also still **added** `m.room.encryption` to new rooms — inconsistent with the backend, and a follow-up failure waiting to happen: with no crypto initialized, the client cannot send events into a room flagged as encrypted.

### Changes

- `src/services/CallManager.ts` — removed all four `assertMatrixRoomEncrypted` guards (call start, group-call permission adjustment, call invite, signalling path); `createElementCallRoom()` no longer sets `m.room.encryption` initial state (comment documents the policy).
- `src/services/matrixCallService.ts` — removed the guards in `startCall` / `answerCall` (1:1 WebRTC path).
- `src/services/matrixGroupCallService.ts` — removed the guard in `startOrJoinGroupCall`.
- `src/services/matrixClientService.ts` — `createDirectMessageRoom()` no longer sets `m.room.encryption` initial state.
- Removed `src/utils/matrixRoomEncryption.ts` and `matrixRoomEncryption.test.ts` — obsolete; no remaining references (verified repo-wide).

No backend, Synapse, or LiveKit changes — those already behave per the new architecture; this aligns the frontend call paths with it.

### Testing

- `eslint --max-warnings=0`, `tsc --noEmit`, `prettier` — clean
- `vitest run src/services` — 33/33 passing


BjoernLudwig:
@shanzaeimran Why was this closed and what issue was addressed?

Issue #411, opened 2026-07-11
[ADR-006-FE] Refresh message timeline after delayed Megolm decryption


Storypapst:
## Problem

A fresh two-consultant PreDev browser run sends a Self-Help TEXT message successfully as `m.room.encrypted` (`200`), and the receiving Matrix Rust-Crypto client later receives the matching Megolm room key, but the visible timeline remains at `Nachricht verschlüsselt`.

## Expected

When a live encrypted Matrix timeline event decrypts after its initial delivery, the open conversation refreshes and renders the clear `m.room.message` without navigation or a manual reload.

## Actual

`SessionStream` refreshes on the initial `Room.timeline` event only. `Event.decrypted` does not flow through `chatTransportService.onMatrixTimeline`, so the already-open timeline keeps the encrypted fallback.

## Evidence

- PreDev: fresh Self-Help TEXT series from the ADR-006 four-modality gate
- Sender: Matrix `PUT .../send/m.room.encrypted/...` returns `200`
- Receiver: Rust-Crypto logs `Received a new megolm room key` for the same room/session
- Receiver UI: new event remains `Nachricht verschlüsselt`
- Reproduces after an additional 45-second pre-send wait

## Acceptance criteria

- A behavior test proves a live `m.room.encrypted` event is surfaced again after successful `Event.decrypted` as `m.room.message`.
- Failed/incomplete decryption attempts do not produce a false clear-message refresh.
- Listener cleanup removes both timeline and per-event decryption listeners.
- The same two-browser PreDev test passes A to B, B to A, and after reload.

Parent gate: #408
Related epic: #410

Storypapst:
PreDev retest after #412 remains RED: sender Matrix event returns 200 and the receiver gets the matching new Megolm room key, but the clear text is not rendered within 60 seconds. The transport now emits the delayed update correctly; the next verified drop point is SessionStream’s 120 ms coalescing branch, which discards rather than trails a second refresh. Reopened for a focused follow-up TDD slice.

Storypapst:
[...]

Issue #408, opened 2026-07-11
[Bug Fix Run] Prove and repair all conversation modalities on PreDev


Storypapst:
Affected repos: ORISO-Frontend, ORISO-UserService, ORISO-Admin, ORISO-TenantService, ORISO-ConsultingTypeService, ORISO-E2E

## Why

Track the complete screenshot-backed Bug Fix Run that started with the ADR-006 four-modality PreDev gate and continued through every regression repair found during the browser run.

## What to build

- Repair browser-proven defects in the owning repositories with regression coverage.
- Preserve successful and failed desktop/mobile evidence in the affected issues.
- Merge through `pre-dev`, deploy regular digest-pinned images, and repeat the repaired browser paths.
- Complete the disposable identity, agency and tenant cleanup through the natural 48-hour safeguard.

## Suggested build order and delivery matrix

| Repository | Native subissue | Pull request | Current state |
| --- | --- | --- | --- |
| ORISO-Frontend | #411 delayed Megolm refresh | prior repair | Closed; evidence attached |
| ORISO-Frontend | #414 null group-chat translations | prior repair | Closed; evidence attached |
| ORISO-UserService | #383 registered-enquiry acceptance | prior repair | Closed; evidence attached |
| ORISO-Frontend | #556 shared modality UI/signalling | #557 | Merged; regular-image recheck pending |
| ORISO-UserService | #514 Live Chat routing and identity | #515 | Merged; regular-image recheck pending |
| ORISO-Admin | #411 agency setup and card width | #412 | Merged; Hot-PreDev evidence attached |
| ORISO-TenantService | #93 with nested #94 | #95 | Merged, deployed and verified |
| ORISO-TenantService | #96 token-bound DPA preview | #97 | In review; CI green |
| ORISO-Frontend | #561 DPA display and encrypted handover | #562 | In review |
| ORISO-UserService | #518 membership and deletion cascades | #519 | In review |
| ORISO-UserService | #530 DPA forwarding | #519 | In review |
| ORISO-ConsultingTypeService | #70 global SMTP delivery | #71 | In review |
| ORISO-Admin | #425 DPA delivery and deletion guards | #426 | In review |
| ORISO-UserService | #531 hidden open enquiry cleanup deadlock | none yet | In progress; natural hard-delete observation pending |
| ORISO-E2E | #14 evidence publication | #15 | PR closed; evidence moves to product issues |

## Verification summary

[...]

PR #413, opened 2026-07-11
fix: retain decrypted Matrix refreshes

Storypapst:
## Problem

The deployed delayed-decryption transport fix correctly re-emits a clear Matrix event, but `SessionStream` can still discard it when it arrives within 120 ms of the encrypted event. The UI then keeps rendering `Nachricht verschlüsselt` even though Rust-Crypto received the Megolm room key.

## Solution

Keep coalescing encrypted/metadata bursts, but never discard a clear `m.room.message`. A component test drives encrypted and decrypted callbacks through one active timeline subscription inside the coalescing window and requires two timeline reads.

## TDD evidence

- Corrected the timeline mock so detach removes stale callbacks.
- Red: expected three total room reads (initial plus encrypted plus decrypted), received two.
- Rejected a trailing-timer implementation because the immediate refresh re-renders and tears down the effect, cancelling the timer.
- Green: clear message events bypass coalescing; focused SessionStream suite 5/5.

## Validation

- Full unit suite: 591/591 passed.
- ESLint + TypeScript: passed.
- Production build: passed with the existing Autoprefixer warnings.
- Final two-browser PreDev A-to-B, B-to-A, and reload proof follows after merge/deploy.

## Scope

Two Frontend files only. No message content leaves the Matrix privacy boundary.

Related bug: #411
Related gate: #408

### DIFF
--- a/src/components/session/SessionStream.test.tsx
+++ b/src/components/session/SessionStream.test.tsx
@@ -24,7 +24,13 @@ const mocks = vi.hoisted(() => {
 		navigate: vi.fn(),
 		logout: vi.fn(),
 		lifecycleListeners: [] as ((change: any) => void)[],
+		timelineListeners: [] as ((
+			event: any,
+			room: any,
+			toStart: boolean
+		) => void)[],
 		detachLifecycle: vi.fn(),
+		getMatrixRoomMessages: vi.fn(() => []),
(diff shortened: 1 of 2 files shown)


Storypapst:
@coderabbitai review

PR #418, opened 2026-07-11
fix: repoint consultant statistics tab to UserService, drop sourceless metrics

Storypapst:
## Why

The consultant profile statistics tab (`profile.routes.activities.statistics`, shown when `featureStatisticsEnabled` is set **or** in non-tenant contexts) calls `/service/statistics/consultant`. That endpoint belonged to the upstream Caritas **statisticsservice, which does not exist in ORISO** — the call 404ed and the tab always showed zeros.

## What

- `endpoints.consultantStatistics` now points at the new UserService endpoint `/service/users/statistics/consultant` (OpenResilienceInitiative/ORISO-UserService#391)
- The tab shows the two metrics the app-layer DB can answer (KDG: own numbers only, computed server-side from the authenticated consultant):
  - **assigned sessions** (created in the selected period)
  - **active sessions** (in progress, touched in the period)
- **Sent messages, video call duration and appointments are dropped** from the DTO, the tiles and the CSV export — they have no app-layer data source in ORISO
- Existing i18n labels are reused (`numberOfAssignedSessions`, `numberOfSessionsWhereConsultantWasActive`), so all languages stay covered with no new keys

## Dependency

Needs OpenResilienceInitiative/ORISO-UserService#391 deployed; until then the tab keeps failing soft (zeros), exactly as today — no regression if this merges first.

## Tests

- `npx tsc --noEmit` clean
- `npm run test:unit`: 72 files / 459 tests green
- Prettier + ESLint clean on the touched files

### DIFF
--- a/src/api/apiGetConsultantStatistics.ts
+++ b/src/api/apiGetConsultantStatistics.ts
@@ -5,10 +5,7 @@ export interface ConsultantStatisticsDTO {
 	startDate: string;
 	endDate: string;
 	numberOfAssignedSessions: number;
-	numberOfSentMessages: number;
-	numberOfSessionsWhereConsultantWasActive: number;
-	videoCallDuration: number;
-	numberOfAppointments: number;
+	numberOfActiveSessions: number;
 }
 
 export interface ApiGetConsultantStatisticsInterface {
--- a/src/components/profile/ConsultantStatistics.tsx
+++ b/src/components/profile/ConsultantStatistics.tsx
@@ -17,7 +17,6 @@ import dayjs from 'dayjs';
 import './statistics.styles';
 import './profile.styles';
 import { useTranslation } from 'react-i18next';
-import { getTenantSettings } from '../../utils/tenantSettingsHelper';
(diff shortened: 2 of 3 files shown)


shanzaeimran (commented):
The fix itself is clean and internally consistent — the slimmed DTO, CSV headers,
CSV data mapping, and render all line up, the now-unused getTenantSettings /
featureAppointmentsEnabled path is fully removed, and the repurposed card's value
and label agree. No blocking bugs.

Two things I'd confirm before merge:

1. Base branch: this targets `dev`, but the other recent feature PRs were
   relocated onto `pre-dev` per the "features go through pre-dev first" policy.
   Should this retarget `pre-dev` for consistency?
2. Backend field names: `numberOfActiveSessions` / `numberOfAssignedSessions` must
   match UserService #391's response exactly, otherwise the tab silently shows 0
   (the .catch swallows errors). Please confirm against the #391 schema.

Non-blocking: the active-sessions card still uses the sent-messages speech-bubble
icon — consider a more fitting icon.

PR #429, opened 2026-07-12
fix: refresh anonymous live-chat stream after send

### LINKED ISSUE #428: Refresh anonymous live-chat stream after late Matrix room join
## Problem

An anonymous live-chat page can load before a consultant accepts the session. The initial session response then has no `matrixRoomId`. The first message now resolves and joins the late-created Matrix room, but the open `SessionStream` still watches the stale session and does not render subsequent consultant messages until a full page reload.

## Scope

- Refresh the active session after the first successful Matrix send when the page loaded without a client room ID.
- Let the existing `SessionStream` rebind its timeline listener to the newly created room.
- Do not reload already-current Matrix sessions or non-Matrix sessions.

## Acceptance criteria

- Anonymous and consultant messages are sent and rendered live without a browser reload.
- Matrix E2EE remains enabled.
- Regression tests cover the refresh decision and callback.
- Merge target is `pre-dev` only.

## Evidence

The current PreDev E2E run records both encrypted Matrix sends as HTTP 200. The consultant sees both messages; the anonymous page times out waiting for the consultant reply because its stream never attached to the late-created room.

Storypapst:
## Summary

- refresh the active session after a successful Matrix send when the page loaded before the room existed
- let `SessionStream` bind its timeline and lifecycle listeners to the late-created anonymous live-chat room
- share the refresh path across text and attachment send success
- keep current Matrix and non-Matrix sessions free from unnecessary reloads

Closes #428

## Tests

- `npm run test:unit -- --run` — 111 files, 662 tests passed
- `npm run lint:scripts` — ESLint and TypeScript passed
- local CodeRabbit review: both findings addressed (no-op branch coverage and shared attachment success path); final rerun hit the temporary plan rate limit

## Release scope

Merge target: `pre-dev` only. Do not merge or push to `dev`.

### DIFF
--- a/src/components/messageSubmitInterface/messageSubmitInterfaceComponent.tsx
+++ b/src/components/messageSubmitInterface/messageSubmitInterfaceComponent.tsx
@@ -25,6 +25,7 @@ import {
 } from './inputField/composerResize';
 import { isAskerEnquirySubmission } from './messageEncryptionMode';
 import { resolveAsideTargetRoomId } from './asideRouting';
+import { reloadSessionAfterSendIfNeeded } from './sessionRefreshAfterSend';
 import { chatTransportService } from '../../services/chatTransportService';
 import { SESSION_LIST_TYPES } from '../session/sessionHelpers';
 import { getModality, Modality } from '../session/getModality';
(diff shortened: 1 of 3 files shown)


Storypapst:
@coderabbitai review

PR #432, opened 2026-07-12
rescue: reduce production Matrix console logging (dev → pre-dev)

kiodreambau:
## rescue

Moves a feature that had landed **directly on `dev`** over to **`pre-dev`**, where it was still missing (per the team policy that features go through `pre-dev`).

### What this is
The Matrix production-logging fix (originally PR #397, commit `5b539e1` by Nikunj Rohit): `matrix-js-sdk` child loggers default to `DEBUG`, so `FetchHttpApi` sync lines flood the production console even after app `console.log` calls are removed. This sets the SDK logger to `error` at startup and makes child namespaces inherit that level.

### Assessment (dev vs. pre-dev)
Comparing `dev` and `pre-dev`, only **one** real feature was unique to `dev` — this one (`src/utils/matrixLogging.ts` did not exist on `pre-dev`). The other two `dev`-only commits were merge commits, and `pre-dev` is otherwise ahead of `dev` (18 commits). This PR is that single rescued feature.

### ⚠️ Bundled scope (heads-up per review)
The rescued commit `5b539e1` (PR #397) is **one commit that bundles several changes** beyond logging. Kept intact here to preserve a faithful rescue rather than re-splitting a colleague's merged commit. What's inside:
- **Matrix logging** (the headline change): `matrixLogging.ts`, `configureMatrixLogging.ts`, `getMatrixAccessToken.ts`, `matrixRegistrationService.ts`, `initApp.tsx`, `polyfill.ts`.
- **Service-origin / dev CORS** (`runtimeConfig.ts` → `resolveServiceOriginForEnvironment`): forces non-local service origins to relative `''` in `NODE_ENV=development`. Dev-only; prod unaffected. (See reply thread — this matches `getApiBaseUrl.ts`'s existing behavior.)
- **Message-render fallback** (`MessageItemComponent.tsx`): `try/catch` around markdown render; now logs at `debug` before falling back to sanitized raw text.
- **`linkify` removal** (`richtextHelpers.ts`): bare-URL linkification is still handled downstream by `urlifyLinksInText` (see reply thread).

### Changes
- `src/utils/matrixLogging.ts` (+ test), `src/configureMatrixLogging.ts`, `src/resources/scripts/runtimeConfig.ts`, `src/polyfill.ts`, `src/initApp.tsx`
- `src/components/message/MessageItemComponent.tsx`, `src/components/sessionCookie/getMatrixAccessToken.ts` (+ test), `src/services/matrixRegistrationService.ts`, and related
- 15 files. Original author preserved via cherry-pick; one follow-up commit adds the debug log in the fallback catch (review feedback).

### How it was rescued
Cherry-picked `5b539e1` onto `pre-dev`. Only trivial conflicts in `.gitignore` (kept the `*.pem` superset) and `.learnings/LEARNINGS.md` (union — kept both sets of notes); all source files auto-merged.

### Testing
`vitest run` on the affected tests — **17 passed** (`matrixLogging.test.ts`, `endpoints.test.ts`, `getMatrixAccessToken.test.ts`).

> Note: the logging code is unchanged from the already-reviewed & merged PR #397, re-homed onto `pre-dev`.

### DIFF
--- a/src/components/messageSubmitInterface/richtextHelpers.ts
+++ b/src/components/messageSubmitInterface/richtextHelpers.ts
@@ -105,8 +105,7 @@ export const markdownToDraftDefaultOptions = {
 	remarkablePreset: 'commonmark',
 	remarkableOptions: {
 		html: true,
-		breaks: true,
-		linkify: true
+		breaks: true
 	}
 };
 
--- a/src/resources/scripts/runtimeConfig.ts
+++ b/src/resources/scripts/runtimeConfig.ts
@@ -150,8 +150,34 @@ export const getMatrixHomeserverUrl = (): string =>
 export const getRuntimeApiBaseUrl = (): string =>
 	ensureHttps(pickValue('REACT_APP_API_URL', 'VITE_API_URL'));
 
+const isLocalServiceOrigin = (value?: string | null): boolean =>
+	/^(https?:\/\/)?(localhost|127\.0\.0\.1)(:\d+)?(\/|$)/i.test(
+		String(value || '').trim()
+	);
+
+/**
+ * In local dev, remote service origins must stay same-origin so webpack-dev-server
+ * can proxy /service and /auth to REACT_APP_DEV_REMOTE_API_URL. Absolute remote
+ * URLs bypass the proxy and trigger browser CORS failures.
+ */
+const resolveServiceOriginForEnvironment = (origin: string): string => {
+	if (!origin) {
+		return '';
+	}
+
+	const nodeEnv =
+		typeof process !== 'undefined' ? process.env.NODE_ENV : undefined;
+	if (nodeEnv === 'development' && !isLocalServiceOrigin(origin)) {
+		return '';
+	}
+
+	return origin;
+};
+
 const getServiceOrigin = (key: string, fallbackOrigin: string): string =>
-	stripTrailingSlashes(ensureHttps(pickValue(key) || fallbackOrigin));
+	resolveServiceOriginForEnvironment(
+		stripTrailingSlashes(ensureHttps(pickValue(key) || fallbackOrigin))
+	);
 
 export const getUserServiceOrigin = (
 	fallbackOrigin = getRuntimeApiBaseUrl()
--- a/.learnings/LEARNINGS.md
+++ b/.learnings/LEARNINGS.md
@@ -15,3 +15,7 @@ TipTap empty docs are markup like `<p></p>`. Using `!composerText.trim()` for �
 ## 2026-07-12 — Browser shortcut conflicts need capture phase
 
 App handlers for Cmd/Ctrl+F, Cmd/Ctrl+Shift+N, Cmd/Ctrl+K must listen in the capture phase and `preventDefault()` before the browser’s default Find / New Window / etc. Bubble-phase listeners are too late. Also set `activeInTextInput: true` when the action must work with the composer focused.
+
+## 2026-07-08 — matrix-js-sdk production logging
+
+`matrix-js-sdk` defaults child loggers to `DEBUG`, so `FetchHttpApi` sync lines appear even when app `console.log` calls are removed. Call `logger.setLevel('error')` at startup and patch `getChild` so child namespaces inherit the same level; pass `logger` into every `createClient` call.
--- a/src/components/agencyRadioSelect/AgencyRadioSelect.tsx
+++ b/src/components/agencyRadioSelect/AgencyRadioSelect.tsx
@@ -29,10 +29,7 @@ export const AgencyRadioSelect = ({
 	const agencyIdAsString = agency.id.toString();
 
 	return (
-		<div 
-			className="agencyRadioSelect__wrapper"
-			onClick={() => console.log('🟣 Wrapper clicked!', agency.id)}
-		>
+		<div className="agencyRadioSelect__wrapper">
 			{prefix && (
 				<Headline semanticLevel="4" styleLevel="5" text={prefix} />
 			)}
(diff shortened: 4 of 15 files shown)


src/resources/scripts/runtimeConfig.ts
```diff
@@ -150,8 +150,34 @@ export const getMatrixHomeserverUrl = (): string =>
...
+	/^(https?:\/\/)?(localhost|127\.0\.0\.1)(:\d+)?(\/|$)/i.test(
+		String(value || '').trim()
+	);
+
+/**
+ * In local dev, remote service origins must stay same-origin so webpack-dev-server
+ * can proxy /service and /auth to REACT_APP_DEV_REMOTE_API_URL. Absolute remote
+ * URLs bypass the proxy and trigger browser CORS failures.
+ */
+const resolveServiceOriginForEnvironment = (origin: string): string => {
```
shanzaeimran, line 163: This dev-origin logic diverges from the convention the codebase already
establishes in `getApiBaseUrl.ts:38-50`. There, development mode honors the
direct-API flag (`VITE_USE_API_URL` / `REACT_APP_USE_API_URL`): if set, return
the remote URL; otherwise return `''` for the proxy. This function only checks
`NODE_ENV === 'development'` and unconditionally forces non-local origins to `''`.

Result: in the supported "call the remote API directly" dev mode
(`VITE_USE_API_URL=true`), `getApiBaseUrl()` returns the remote host, but
`getServiceOrigin()` — which backs user-service, service-settings, and
keycloak/auth (per endpoints.test.ts) — is still forced to relative `''`. The
base API and the service/auth origins then point at different places and those
requests break. Prod is unaffected, so this is dev-only, but it reimplements a
decision that's already centralized.

Suggest reusing the existing `isDirectApiUrlEnabled(processEnv)` check so both
paths agree, rather than a standalone `NODE_ENV` branch here.

src/components/messageSubmitInterface/richtextHelpers.ts
```diff
@@ -105,8 +105,7 @@ export const markdownToDraftDefaultOptions = {
 	remarkablePreset: 'commonmark',
 	remarkableOptions: {
 		html: true,
-		breaks: true,
-		linkify: true
```
shanzaeimran, line 109: Low risk but worth a QA note since it touches every rendered message: this
options object is consumed only in received-message rendering
(MessageItemComponent.tsx:866), where `urlifyLinksInText` runs on the output HTML
right after (line 876) and independently linkifies bare URLs. Explicit markdown
`[text](url)` links come from the commonmark parser regardless of `linkify`, so
no regression is expected — please just confirm bare-URL rendering in a message
still linkifies as before.

shanzaeimran (commented):
The core change — silencing matrix-js-sdk's default DEBUG child loggers and
wiring the logger into every createClient call — is clean, correct, and well
tested. No concerns there.

One thing I'd fix before merge, plus a scope note:

1. `runtimeConfig.ts`'s new `resolveServiceOriginForEnvironment` ignores the
   `VITE_USE_API_URL` / `REACT_APP_USE_API_URL` direct-API dev flag that
   `getApiBaseUrl.ts` honors. In that dev mode the base API points remote while
   service/auth origins are forced relative, breaking those requests. Reuse
   `isDirectApiUrlEnabled` so both agree. (Dev-only; prod unaffected.)

2. Scope: this PR is titled/described as a Matrix console-logging rescue, but it
   also bundles the runtimeConfig CORS change above, a message-render try/catch,
   and the `linkify` removal. The service-origin change is the one thing here
   that can actually break requests, and it's easy to miss under a "logging"
   title — consider splitting it out or calling it out explicitly in the
   description.

Non-blocking: confirm bare-URL linkification still works after dropping
`linkify`, and consider logging (at debug) in the new empty catch.

kiodreambau:
Thanks @shanzaeimran — addressed each point:

**1. `resolveServiceOriginForEnvironment` vs. `getApiBaseUrl` (direct-API dev flag).**
I checked this closely and I don't think a change is warranted, because the premise doesn't hold for this codebase:

- `getApiBaseUrl.ts` does **not** honor a `VITE_USE_API_URL` / `REACT_APP_USE_API_URL` flag. In `NODE_ENV=development` it returns the remote URL **only** when `buildTimeApiUrl` is a *local* host (`isLocalApiHost`), and `''` otherwise:
  ```ts
  if (processEnv.NODE_ENV === 'development') {
      if (buildTimeApiUrl && isLocalApiHost(buildTimeApiUrl)) return normalizeApiUrl(buildTimeApiUrl);
      return '';
  }
  ```
- `resolveServiceOriginForEnvironment` applies the **same** rule: non-local origins → `''` in dev, local origins kept (`isLocalServiceOrigin`).
- `isDirectApiUrlEnabled` / `USE_API_URL` don't exist anywhere in the repo — I grep'd `dev`, `pre-dev`, `main` and history (0 hits). So there's no supported dev mode where the base API points at a remote host while the service origin is forced relative; both functions force non-local → `''` and keep local, so they stay in agreement.

If you were looking at a different branch/project where that flag exists, happy to reconcile — but on `dev`/`pre-dev` the two are already consistent, so I've left the code unchanged.

**2. Scope.** Fair point — I've expanded the PR description with an explicit **"Bundled scope"** section listing the four bundled changes (logging, the `runtimeConfig` service-origin/CORS change, the message-render `try/catch`, and the `linkify` removal). Kept the commit intact rather than re-splitting a colleague's already-merged `#397`.

**3. Non-blocking.**
- **Bare-URL linkification confirmed (by code):** `linkify` only affected the commonmark parse step; the rendered output still runs through `urlifyLinksInText(stateToHTML(...))` at `MessageItemComponent.tsx:876`, which independently linkifies bare URLs, and explicit `[text](url)` links come from commonmark regardless. So no regression is expected.
- **Empty catch:** added a `console.debug(...)` in the markdown-render fallback (commit `6377a75`), consistent with this PR's goal of an intentional/quieter console.

---
_Generated by [Claude Code](https://claude.ai/code)_

shanzaeimran:
Thanks — you're right on point 1, withdrawing it. I was comparing against a version of `getApiBaseUrl.ts` with a `VITE_USE_API_URL` escape hatch that doesn't exist on this branch. Here both functions apply identical dev logic (non-local →
`''`, local kept), so they stay in agreement. My mistake.

The `Bundled scope` section addresses point 2, and point 3 is covered — thanks for confirming `urlifyLinksInText` and adding the debug log in `6377a75`.

No blockers from me. LGTM 👍

Issue #435, opened 2026-07-13
Relations foundation: replace reply with m.relates_to, real threads (MSC3440), editing, reactions, intentional mentions


kiodreambau:
Part of the Element-adoption epic. Foundation block — **do this first**: reply, threads, editing, reactions and mentions all ride the same Matrix relations mechanic (`m.relates_to`). Our current reply implementation is broken and gets **replaced by the relations model, not patched**.

Reference implementation to study (AGPL — patterns only): `~/Documents/GitHub/element-web/apps/web/src` (ReplyChain, threads hooks incl. `useRoomThreadNotifications`, autocomplete provider architecture). The relation semantics themselves are Matrix spec, SDK support (matrix-js-sdk, Apache-2.0) is complete and usable directly.

## Tasks

**Reply (fix by replacement)**
- [x] Model replies as `m.relates_to` / `m.in_reply_to` relations on the event (drop the current custom mechanism) *(PR #445)*
- [x] Fallback stripping for legacy clients *(PR #445)*
- [x] Reply-chain rendering in the message list *(PR #445)*
- [x] Reply preview in the TipTap composer (cancelable) *(PR #445)*

**Threads (MSC3440)**
- [x] `m.thread` relations via SDK thread support *(PR #445 — relation-write + dual-read with legacy [THREAD:] prefix fallback)*
- [x] Thread panel + per-room thread list *(PR #445 — "Threads (N)" panel listing every thread, root preview + reply count + unread)*
- [x] Thread summaries on the root message (last reply + counter) *(PR #445)*
- [x] Per-thread unread state *(PR #445 — device-local last-read timestamp per room+thread; NOT full MSC3771/3773 cross-device receipts, flagged as a simplification pending live-homeserver verification)*
- [x] Replace the URL-only `threadRootId` plumbing with the real data model *(PR #445 — threadRootEventId from the relation drives summaries/filtering/notifications)*

**Editing & reactions (same mechanic, one pass)**
- [x] Message editing via `m.replace` (edit history preserved) *(PR #445 — `applyMessageEdits` folds edits onto the original, edited marker, edit-in-place via composer + menu action)*
- [x] Reactions via `m.annotation` + aggregation rendering *(PR #445 — aggregated pills + quick-react picker, live-updates via timeline listener)*

**Intentional mentions**
[...]

Issue #434, opened 2026-07-13
Epic: Element-grade chat foundation — relations, crypto, settings, room organisation


kiodreambau:
Umbrella for adopting proven Element Web technology into the ORISO chat, per the 2026-07-13 planning session. The full analysis lives in the vault: *7 Work Packages/WP-06 Activity Timeline/Element Web Notification System — Adoption Analysis.md*; local source clone for study: `~/Documents/GitHub/element-web` (upstream `develop`, 2026-07-10 + tag v1.11.55).

**License note:** all cryptography lives in **matrix-js-sdk (Apache-2.0)** — we use it directly. Only the Element *UI* on top is AGPL and must be pattern-reimplemented, never copied.

## The five blocks (recommended order)

1. **Relations foundation** — reply fix via `m.relates_to`, real threads (MSC3440), editing (`m.replace`) & reactions (`m.annotation`), intentional mentions (`m.mentions`). Four features, one mechanic.
2. **Crypto track** (parallel) — key backup + recovery UX, invisible crypto (MSC4153), dehydrated devices (MSC3814), UTD telemetry. Inherits Rust crypto (vodozemac) from the SDK for free (ADR-012 Megolm migration).
3. **Unified settings store** — Element's level model (device → room-device → room-account → account → config) to unify releaseToggles / AppConfig / profile endpoints / localStorage.
4. **Room organisation** — tags, pinned messages, read markers/receipts, permalinks.
5. **Notifications completion** — tracked in #420 (Slices 6a/6b landed via #419; per-conversation 4-state menu, real preview component, @mentions chip remain).

Deliberately out of scope: server-side full-text search in E2EE rooms (impossible by design — confirms ADR-AT-01 client-side filtering), link previews (privacy: homeserver fetches URLs — off by default in a counselling context).

Cross-cutting deps: backend event writers (WP-06 Slices 2–5) live in ORISO-UserService and are tracked separately.

kiodreambau:
Progress check: 1 of 6 native sub-issues closed (16%). Sub-issue #435 (relations foundation) has a merged PR (#445, "slices 1–4") but the sub-issue itself remains open — reactions/edits/threads/mentions are only partially landed. #436 (crypto), #441 (settings), #442 (room organisation), and #420 (activity timeline) show no linked PRs yet. Concrete next step: close out the remaining relations slices on #435, then start the crypto track (#436).

Storypapst:
[...]

Issue #420, opened 2026-07-12
Activity Timeline: Figma icon set, real chat preview, mention filter, Element-style notification settings


kiodreambau:
Follow-ups from the 2026-07-12 design review of the Activity Timeline (WP-06), on top of #419. Figma sources: App.Oriso → *Zeitstrahl | Timeline* page (frames `7576-22889`, `7582-27615`) and Design-System M3 → *Timeline History Pannel* (`61055-49256`).

## UI polish
- [x] **Replace the interim icon mix with the Figma icon set.** *(Done in PR #419 via Material Symbols — the M3 kit's canonical set; exact Figma exports can still be swapped in via `eventDescriptors/icons.tsx` if design provides them.)* Cards and family chips currently use a mix of MUI icons and repo SVGs; several don't match the Figma per-event icons. Design will provide exported SVGs from the Design-System M3 file; wire them through `eventDescriptors/icons.tsx` (icon ids already exist per event type).
- [ ] **Migrate `SessionsListToolbar` to the shared `ListSearchField`** molecule introduced in #419 so both master lists render the identical primitive.
- [ ] **Appointments chip**: family is reserved in the registry; appears automatically once appointment events are emitted (backend WP-06 out-of-scope note) — verify chip + icon then.

## Preview
- [ ] **Real embedded conversation preview** instead of the `embeddedNotifications=1` iframe: render the session view component directly in the detail pane (Slack-style activity preview). E2EE-safe by construction — same user, same client, decryption stays client-side (ADR-AT-01). The iframe workaround also breaks in Storybook.

## Filters
- [ ] **@Mentions filter chip**: filter to events where the user was mentioned, composing with the search field (e.g. mentions + person query). Requires mention detection — follow Element's push-rule content-rule pattern (`is_user_mention`/keyword rules) rather than a bespoke parser.

## Element-pattern adoption (see vault: *WP-06 Activity Timeline/Element Web Notification System — Adoption Analysis*)
- [x] **Slice 6a — settings transport** *(Done in PR #419: account-data settings + MSC3890 device silence + suppression gate; decrypt-then-notify queue remains open for the Megolm track.)*: notification preferences as Matrix push rules (cross-device for free); per-device 'silence this browser' via MSC3890-style device-scoped account data; decrypt-then-notify queue as the Megolm gate for browser notifications.
[...]

PR #445, opened 2026-07-13
feat(chat): relations foundation — reply, threads, editing, reactions (#435, slices 1–4)

kiodreambau:
Relations foundation #435 (epic #434), base `pre-dev`.

## What

'Reply directly' existed in the message action menu but had **no handler** — reply was effectively dead, and threads were a `[THREAD:rootId]` text-prefix hack. This PR replaces both with the Matrix spec relation mechanic, then extends the same mechanic to editing and reactions.

### Slices 1–2: Reply (`m.in_reply_to`) + Threads (`m.thread`)

- **`messageRelations` pure helpers**: build `m.relates_to`/`m.in_reply_to` content, read the relation (incl. the MSC3440 thread-fallback shape), strip the legacy Element quote fallback.
- **Transport**: `replyToEventId`/`threadRootId` flow composer → `apiSendMessage` → `chatTransportService` → Matrix event content. The metadata event notification stays text-free — a test pins the FE-H01 boundary (no message text, no event ids in the payload).
- **Timeline**: formatter exposes `replyToEventId`/`threadRootEventId` + strips the fallback; message cards render a quote block, resolved from the loaded timeline.
- **Composer**: cancelable reply-quote preview docked inside the composer box; menu entry only renders where a handler is wired (main timeline, not the thread panel).

### Slices 3–4: Editing (`m.replace`) + Reactions (`m.annotation`)

- **Editing**: `buildEditContent`/`getReplaceTargetId`/`getEditedBody` helpers; `applyMessageEdits` folds edit events onto the original message (drops the edit event, overwrites the body with the latest edit by timestamp, marks `isEdited`). Transport: `matrixClientService.editMessage` (retry-on-expired-token, same pattern as `sendMessage`) → `chatTransportService.editTextMessage`. Composer gets an `editingMessage` prop with a cancelable edit-in-progress banner and prefill; "Edit Message" menu item on own messages. Edited marker renders next to the bubble timestamp.
- **Reactions**: `buildReactionContent`/`getReactionTarget`/`aggregateReactions` helpers; raw `m.reaction` events are collected separately from the formatted message list (`extractReactionEvents`, since reactions are a distinct event type) and aggregated per message by emoji key. Transport: `matrixClientService.sendReaction`/`redactEvent` → `chatTransportService.sendReaction`/`removeReaction`. UI: aggregated pills below the message bubble (own reaction highlighted, click to un-react) + a quick-react picker (6 emoji). Live timeline listener now also refreshes on `m.reaction`/`m.room.redaction` so reactions update for other participants without a reload.

### Merge note

This branch was rebased against `pre-dev` mid-session, which had independently picked up an unrelated, complementary edit feature (keyboard-shortcut "edit last own message", `editingMessageId`/`onLocalMessageEdit` in the composer, from the self-help-group-chat work). Both coexist post-merge — different trigger paths (menu click on any own message vs. shortcut on your last message), both call into `matrixClientService.editMessage`. They don't currently share state, so triggering both without cancelling one first is a latent (narrow) edge case; flagging for a follow-up unification pass rather than blocking on it here.

## Verification

- **69 new vitest cases** across `messageRelations`, `matrixTimelineEventFormatter`, `matrixClientService`, `chatTransportService`, `sessionHelpers` — TDD (red→green) throughout.
- Full suite after merge: **770/776 passing**; the 6 failures are pre-existing, unrelated to this branch (`notificationSettings.test.ts` + `StageLayout.test.tsx`, a `localStorage` test-isolation issue not touched by this PR).
- `tsc --noEmit` clean.
- **Not verified this session**: Storybook visual check of the new editing/reactions UI — the local Storybook dev server repeatedly failed to stay up in this environment (process reaped mid-boot across 4 attempts over ~8 minutes) despite the `EditingMessage` story being added (`MessageSubmitInterface.stories.tsx`). The reply-preview UI it mirrors was visually verified in slice 1; reaction pills have no story coverage yet (`MessageItemComponent` isn't storied — heavy context-provider dependency).
- E2E send/receive against a live homeserver is **not** covered here (Pre-Dev E2E accounts still missing).

Closes checklist items "Message editing via `m.replace`" and "Reactions via `m.annotation` + aggregation rendering" on #435.


kiodreambau:
**Slice 2 pushed** (`59bd6c6`): thread membership is now the MSC3440 `m.thread` relation instead of the `[THREAD:rootId]` plaintext prefix inside the message body. Send side writes the relation (incl. reply-inside-thread with `is_falling_back: false`); read side is dual — relation first, legacy prefix as read-only fallback, so existing conversations keep rendering. 226/226 green, tsc clean.

kiodreambau:
**Slice 2b added: Thread panel + per-thread unread**

- **Thread list panel**: "Threads (N)" toggle in the session content area lists every thread in the room (root preview, reply count, unread indicator), clicking an entry opens it — same `handleOpenThread` as the existing thread panel. `threadSummaries` computation extracted from an inline `useMemo` into a pure, unit-tested helper (`src/utils/threadSummaries.ts`).
- **Per-thread unread**: `src/utils/threadUnread.ts` — a device-local last-read timestamp per (room, thread root), localStorage-backed. This is explicitly **not** full cross-device Matrix thread read receipts (MSC3771/3773); that needs a live homeserver to verify against threads, which isn't available in this environment. Flagging as a known simplification, not a spec implementation.
- **Found + fixed while TDD-ing the unread store**: `jsdom@24.1.3` (pinned in `package.json`) doesn't implement the Web Storage API at all — every jsdom-environment test touching `localStorage` threw, including two pre-existing failures unrelated to this branch (`notificationSettings.test.ts`, `StageLayout.test.tsx`/`DevToolbar`). Added a minimal in-memory `Storage` polyfill to the existing `src/test/jsdomPolyfills.ts` setup file. Full suite is now **786/786 green** — first fully clean run this session (was 656/662, then 786/6-failing after the merge, now 786/786).

Closes the remaining two "Threads (MSC3440)" checkboxes on #435.

kiodreambau:
**Slice 3 added: Intentional mentions (\`m.mentions\`)**

Completes #435's mentions checklist. The TipTap autocomplete + mention-pill UI already existed from earlier work (\`createMentionExtension\`, \`MentionList\`, \`mentionFiltering\`) — the extension's own comment flagged that it "stores the consultant id + matrix id so the message layer can serialize m.mentions," but nothing did yet. This closes that gap:

- **\`src/utils/messageMentions.ts\`**: \`extractMentionedUserIds\` (reads \`data-mention-matrix-id\` off composer HTML), \`buildMentionsContent\` (the \`m.mentions\` content fragment, merged into \`buildTextMessageContent\` alongside the existing reply/thread relations), \`getMentionedUserIdsFromContent\` (read side).
- **Robust + E2EE-safe by construction**: a mention pill only ever carries a *resolved* Matrix room-member id — matched via the existing \`audienceOptions\` room-member resolution (reused, not duplicated), never a display name or a guessed id. A consultant not yet in the room simply can't be mentioned in \`m.mentions\` (the pill still shows visually, flagged "not in chat", consistent with existing UX) — no display-name matching anywhere in the wire-format path.
- **TipTap**: extended the \`Mention\` node with a \`matrixUserId\` attribute (\`.extend({ addAttributes() {...} })\`), threaded through \`MentionList\`'s \`command()\`.
- **Read side**: \`formatMatrixTimelineEvent\` now exposes \`mentionedUserIds\` from incoming \`m.mentions\` — ready for the timeline @mentions filter chip (#420) to consume; that chip's UI itself stays in #420's scope.

30 new vitest cases (TDD, red→green). Full suite: **802/802 green**, \`tsc\` clean.

Closes the remaining "Intentional mentions" checklist block on #435 for the m.mentions/autocomplete/pill items; the timeline filter chip stays tracked in #420 as noted there.

Storypapst:
Requested improvement before merge — **ADR-017** (`0 - Docs/ADR-017-native-matrix-threads-hard-cut.md`, decided 2026-07-18): threads move to native `m.thread` **with a hard cut**, no dual-read. The relations work in this PR is exactly right; only the legacy `[THREAD:]` compatibility path needs to come out:

**1. Remove the legacy fallback reads (relation is the only source of thread identity):**
- `src/utils/threadSummaries.ts` — `const rootId = message.threadRootEventId || parsed.threadRootId;` → use `message.threadRootEventId` only (and drop the file's "legacy-[THREAD:]-prefix-fallback" contract comment).
- `src/utils/matrixTimelineEventFormatter.ts` — remove the "falls back to the legacy `[THREAD:rootId]` prefix when there is no relation" branch and its spec in `matrixTimelineEventFormatter.test.ts`.
- Any remaining thread-panel/filter site that still consults `parsed.threadRootId` (e.g. the `renderMode="thread"` filtering path in `SessionItemComponent`/`MessageItemComponent`).

**2. Delete the thread branch of the prefix convention in `src/components/message/messageConstants.ts`:**
- `THREAD_PREFIX`, `THREAD_SUFFIX`, `buildThreadPrefix`, and the `isThreadMessage`/`threadRootId` outputs of `parseMessagePrefixes`.
- ⚠️ Keep the `[SUPERVISOR_FEEDBACK]` / `[VISIBLE_TO:]` parsing untouched — that is ADR-008 aside routing, not thread legacy.

**3. Keep one cosmetic render filter:** `cleanedMessage` should still strip a leading `[THREAD:…]` token from message bodies so leftover pre-dev events don't display the raw token — but it must carry **no thread semantics** (stripping only, nothing returned).

Rationale: house rule "no prod users → don't carry migration burdens" — pre-dev thread data is disposable, and a permanent second read path for disposable data is the exact legacy ADR-017 retires. Old prefix messages simply render flat.

Refs OpenResilienceInitiative/ORISO-Frontend#435 · ADR-017

kiodreambau:
Done — ADR-017 hard cut applied in `4eb4d8d`:

**1. Legacy fallback reads removed** (relation `message.threadRootEventId` is the only thread-identity source):
- `threadSummaries.ts` — dropped `|| parsed.threadRootId` and the legacy-fallback contract comment.
- `SessionItemComponent.tsx` — both new-thread-reply sites read the relation only.
- `MessageItemComponent.tsx` — `effectiveThreadRootId = threadRootEventId || null` (was `|| parsedMessage.threadRootId`).
- ℹ️ `matrixTimelineEventFormatter.ts` already read threads **relation-only** (`getThreadRootId`) — it had no `[THREAD:]` branch; its `stripReplyFallback` is the *reply* quote fallback (ADR-008-adjacent), left untouched. The legacy spec actually lived in `threadSummaries.test.ts`, now flipped to assert the prefix produces **no** summary.

**2. `messageConstants.ts`** — deleted `THREAD_PREFIX`, `THREAD_SUFFIX`, `buildThreadPrefix`, and the `isThreadMessage`/`threadRootId` parse outputs. `[SUPERVISOR_FEEDBACK]` / `[VISIBLE_TO:]` (ADR-008 aside routing) untouched — VISIBLE_TO's closing bracket moved to a neutral `PREFIX_SUFFIX` const.

**3. Cosmetic strip kept** — `cleanedMessage` still strips a leading `[THREAD:…]` token (stripping only, no semantics) so leftover pre-dev events don't render the raw marker. New `messageConstants.test.ts` guards it (incl. combined with VISIBLE_TO).

Old prefix messages now render flat. 494 tests green, tsc clean.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

shanzaeimran (approved):

PR #463, opened 2026-07-14
OBS-P8: browser Real User Monitoring (Core Web Vitals) via SigNoz

Storypapst:
## Summary
- Adds an OpenTelemetry MeterProvider (`src/utils/observability/meterProvider.ts`), registered with `service.name: 'frontend'` and a `PeriodicExportingMetricReader` exporting to the self-hosted SigNoz collector (`https://signoz.oriso-dev.site/v1/metrics`, 10s interval) via `OTLPMetricExporter`.
- Adds Core Web Vitals capture (`src/utils/observability/webVitals.ts`) wiring `web-vitals`' `onCLS`/`onINP`/`onLCP`/`onTTFB`/`onFCP` reporters into OpenTelemetry instruments named exactly `cls` (observable gauge), `inp`/`lcp`/`ttfb`/`fcp` (histograms) on a meter named `web-vitals` — matching the SigNoz dashboard's expected names exactly.
- Attaches only one attribute to every recorded metric: `url.path_template`, computed by `normalizedPagePath()`, which strips the query string and replaces both numeric (`/\d+/`) and UUID-shaped path segments with `:id` (this app's routes carry UUIDs, e.g. session/agency/counsellor ids, not just numeric ones — extended past SigNoz's numeric-only sample regex).
- Wires both into `src/initApp.tsx`, calling `initMeterProvider()` then `initWebVitals()` as early as possible, before anything else can call `metrics.getMeter(...)` (the OTel API snapshots whatever provider is globally registered at call time, so ordering matters).
- Adds `@opentelemetry/api`, `@opentelemetry/sdk-metrics`, `@opentelemetry/exporter-metrics-otlp-http`, `@opentelemetry/resources` as dependencies (`web-vitals` was already a declared, previously-unused dependency at `^3.3.2`, which already exports `onCLS`/`onINP`/`onLCP`/`onTTFB`/`onFCP`/`Metric`, so it was left as-is).

## Deliberately left out
Per the OBS-P8 scope and ADR-011 / the OBS-P6 pseudonymization work: no `user.id` (from localStorage) or browser/user-agent identifier is attached as a metric attribute, even though SigNoz's own docs show that as an optional "attach custom identifiers" variant. This codebase treats SigNoz strictly as developer/ops observability tooling, never a source of per-user tracking. `url.path_template` (already scrubbed of ids) is the only attribute. Could add non-identifying attributes later (e.g. a coarse device-type bucket) if the dashboard needs more slicing, but intentionally not doing that here.

Refs OpenResilienceInitiative/ORISO-Helm#62

## Test plan
- [x] `npm run lint:scripts` (eslint + tsc) — clean
- [x] `npm run test:unit` — 756/756 passing across 120 files, including 21 new tests covering: `normalizedPagePath` (query-string stripping, numeric segments, UUID segments, combined, already-clean paths), exact histogram/gauge names (`lcp`/`cls`/`inp`/`ttfb`/`fcp`) and meter name (`web-vitals`), the `MeterProvider` resource (`service.name: 'frontend'`), the exporter URL/interval, that CLS goes through the observable-gauge callback (not a direct `record`), that no attribute besides `url.path_template` is attached, idempotent init, and that setup/recording never throws even when the SDK is misconfigured or export fails
- [x] `npm run build` (production build) — succeeds, only pre-existing unrelated warnings (autoprefixer CSS, babel-preset-react-app deprecation, bundle size)
- [x] `npm ci --ignore-scripts --legacy-peer-deps` (mirrors CI's install step) — verified clean against the updated lockfile

Not merging — for review only.


Storypapst:
Reviewed: contract matches exactly (meter/histogram names, service.name='frontend', url.path_template-only attribute, no user.id/browser-identity). Good catch on the getMeter()-snapshots-at-call-time hazard and making the init order explicit rather than relying on import order. All 4 CI checks green. Merging.

Issue #464, opened 2026-07-14
Matrix E2EE crypto is never initialized — encrypted messages are never actually decrypted


Storypapst:
## Problem

`client.initRustCrypto(...)` (or any crypto-init call) is never invoked anywhere in `src/`. Repo-wide search for `initRustCrypto`, `initCrypto`, `RustCrypto`, `setupRust`, `getCrypto()`, `isCryptoEnabled` returns zero hits outside `node_modules`.

`createMatrixClient` (`src/components/sessionCookie/getMatrixAccessToken.ts:101-111`) only calls `createClient({ baseUrl, accessToken, userId, deviceId, fallbackICEServerAllowed })`. `initializeClient` (`src/services/matrixClientService.ts:97-136`) only calls `client.startClient(...)`. Neither initializes the SDK's crypto engine.

matrix-js-sdk (`^38.4.0`, pinned in `package.json`) ships **Rust crypto (vodozemac) only** in this version — there is no legacy libolm fallback path. Without an explicit `initRustCrypto()` call, `client.getCrypto()` returns `undefined` and the client never attempts to decrypt anything.

## Impact

Every `m.room.encrypted` event in every room is, and always has been, effectively permanently undecryptable — not due to a key-delivery problem, but because decryption is never attempted at all. `src/utils/matrixTimelineEventFormatter.ts:26-27` treats any encrypted event lacking a `msgtype` as `isUndecryptedEvent` and shows placeholder text — which is exactly what you'd see under this condition, and is consistent with the current UI behavior for encrypted rooms.

This is the load-bearing assumption behind the whole Element-adoption crypto track (#434, #436-440): key backup UX, invisible crypto (MSC4153), dehydrated devices, and decryption-failure (UTD) telemetry (#440) all assume the crypto engine is actually running. None of that work is meaningful until this is fixed.

## Ask

Confirm this is a real gap (not something intentionally deferred that I'm missing context on) and prioritize wiring up `initRustCrypto()` in the client bootstrap path, alongside whatever key-storage/cross-signing setup that requires. This is a substantial, security-relevant change on its own — not something to bolt onto an unrelated task.

Found while researching OBS-P9 (Matrix decryption-failure telemetry, tracked in ORISO-Helm#62) — flagging separately since it's a functional/security gap independent of observability.

Storypapst:
**Correction — this finding was wrong.** Re-checked directly against \`origin/pre-dev\`:

\`\`\`
[...]

Issue #440, opened 2026-07-13
Decryption-failure telemetry (UTD tracker → SigNoz)


kiodreambau:
Crypto-track sub-issue. **Decryption-failure telemetry** ("unable to decrypt", UTD) — the health metric of an E2EE chat. We found the delayed-decryption bug (#412) by luck; after the Megolm cutover we need eyes.

Element's `DecryptionFailureTracker` pattern (AGPL — reimplement; the SDK emits the failure events for free, Apache-2.0):
1. **Grace period** — transient failures (key arrives seconds late — exactly our #412 symptom) self-heal and are not counted
2. **Cause classification** — key never arrived / withheld on purpose (MSC4153!) / historical without backup / wrong Megolm index — each class has a different fix
3. **Aggregated reporting** — for us: **SigNoz/OTel** instead of PostHog

## Tasks
- [x] Reimplement tracker core (grace period + cause classification) on SDK decryption-failure events
- [x] Export aggregated UTD metrics via OTel to SigNoz
- [ ] SigNoz dashboard: UTD rate per cause; alert on post-deploy regression (ops/infra !92 ORISO-Helm#62)
- [x] Wire MSC4153 "intentionally withheld" class separately (expected, not a bug)
- [x] Regression test reproducing the #412 delayed-key scenario against the grace period

kiodreambau:
**Audit (2026-07-15): most of this shipped already.** The tracker landed as commit `aba5fca` (OBS-P9), wired in `initApp.tsx`:
- Task 1 (grace period + cause classification) ✅
- Task 2 (`decryption_failure` counter, OTel → SigNoz via `meterProvider`) ✅
- Task 4 (MSC4153 withheld vs bug `category`) ✅

29 tests were green already (they only appeared red in a stale checkout missing the `@opentelemetry/*` deps — `npm install` fixes it).

**Added now — Task 5:** a named #412 regression guard (delayed key self-heals → `transient_resolved`, never arrives → permanent `bug`). PR #501. → 31 tests.

**Remaining — Task 3** is ops/infra: SigNoz dashboard (UTD rate per cause) + post-deploy regression alert. That belongs to the observability rollout epic **ORISO-Helm#62**, not this repo — flagged there, out of scope for the frontend.

PR #465, opened 2026-07-14
feat: add UTD (decryption-failure) telemetry tracker (OBS-P9)

Storypapst:
## Summary

- Adds `initUtdTracking()` (`src/utils/observability/utdTracker.ts`), wired after `initMeterProvider()`/`initWebVitals()` in `initApp.tsx` -- the OTel-export half of the "Unable To Decrypt" (UTD) tracker for encrypted counselling conversations.
- Observes every room's `Event.decrypted` firings on `m.room.encrypted` timeline events via a client-level `Room.timeline` listener (polling for the Matrix client to become available/change, since the client doesn't exist yet at app startup and can be recreated on token refresh).
- A per-event 5-minute grace-period timer classifies each failure as `transient_resolved` (key arrived before the timer fired, optional/nice-to-have signal) or `permanent` (still failing after 5 minutes, the primary signal). `category` splits `permanent` failures into `withheld` (`MEGOLM_KEY_WITHHELD` / `MEGOLM_KEY_WITHHELD_FOR_UNVERIFIED_DEVICE` -- MSC4153-adjacent, intentional/expected) vs `bug` (every other code).
- **No PII/identifying attributes ever**: only `{outcome, cause}` or `{outcome, cause, category}` are attached -- never room id, event id, sender, or user id. Same dev/ops-only telemetry principle as OBS-P3/OBS-P8 (ADR-011), doubly so here since this observes encrypted counselling conversations specifically.
- Meter name `'utd-tracker'`, counter name `'decryption_failure'` -- exact strings a follow-up SigNoz dashboard will query by name.

## Relation to #412 (`fix: refresh delayed Matrix decryption`)

\#412 is already merged into `pre-dev` (commit `20ad545`) and already implements a 5-minute grace-period pattern for `Event.decrypted` in `chatTransportService.onMatrixTimeline` -- but for a different purpose (UI refresh) and a different scope (only the single room a `SessionStream` component currently has open, via `onMatrixTimeline(matrixRoomId, listener)`).

Since this telemetry needs to span *every* room, not just whichever one is currently open in the UI, #412's per-room mechanism can't be reused as-is. This PR ships its own, separate, minimal grace-period tracker (same 5-minute window, for classification consistency) wired directly onto the Matrix client's `Room.timeline` event, which already fires for all rooms. `chatTransportService.ts` and `meterProvider.ts`/`webVitals.ts` are untouched -- this PR only adds two new files plus the two-line wiring (`initApp.tsx`).

## Metric schema (for the follow-up SigNoz dashboard)

- Meter: `utd-tracker`
- Counter: `decryption_failure`
- Attributes:
  - Permanent failure: `{ outcome: 'permanent', cause: <DecryptionFailureCode>, category: 'withheld' | 'bug' }`
  - Transient resolved (optional/nice-to-have): `{ outcome: 'transient_resolved', cause: <DecryptionFailureCode> }`

**Known limitation, tracked separately, not addressed here**: matrix-js-sdk's crypto engine is never actually initialized anywhere in this app today (#464), so `Event.decrypted` basically never fires for real yet -- this tracker will show near-zero live data until that separate issue is fixed. Built correctly anyway so it's live the moment crypto init lands.

Refs OpenResilienceInitiative/ORISO-Helm#62
Refs OpenResilienceInitiative/ORISO-Frontend#440

## Test plan

- [x] `npx eslint src/utils/observability/utdTracker.ts src/utils/observability/utdTracker.test.ts src/initApp.tsx --max-warnings=0`
- [x] `npx tsc` (full project typecheck)
- [x] `npm run test:unit` -- 121 files / 785 tests pass, including 29 new tests covering: grace-period resolve-before-timeout vs timeout-fires, all 12 `DecryptionFailureCode` values classified into `withheld`/`bug`, exact attribute-set assertions (no PII), meter/counter name exactness, dedupe of repeated `Room.timeline`/`Event.decrypted` fires, client-swap re-attach, and never-throw resilience -- using `vi.useFakeTimers()` to advance past the 5-minute window.
- [x] `CI=false npm run build` -- succeeds (only pre-existing, unrelated warnings)

### DIFF
--- a/src/initApp.tsx
+++ b/src/initApp.tsx
@@ -19,6 +19,7 @@ import { Privacy } from './components/legalInformationLinks/Privacy';
 import { Imprint } from './components/legalInformationLinks/Imprint';
 import { initMeterProvider } from './utils/observability/meterProvider';
 import { initWebVitals } from './utils/observability/webVitals';
+import { initUtdTracking } from './utils/observability/utdTracker';
 
 // OBS-P8 (ORISO-Helm#62): browser-side Real User Monitoring. Register the
 // MeterProvider before anything else gets a chance to call
@@ -28,6 +29,9 @@ import { initWebVitals } from './utils/observability/webVitals';
 // load. Both are best-effort: see the try/catch in each module.
 initMeterProvider();
 initWebVitals();
+// OBS-P9 (ORISO-Helm#62, ORISO-Frontend#440): Unable-To-Decrypt (UTD)
+// failure tracking for encrypted counselling conversations.
+initUtdTracking();
 
 const ThemeDemo = lazy(() =>
 	import('./components/themeDemo/ThemeDemo').then((m) => ({
--- a/src/utils/observability/utdTracker.test.ts
+++ b/src/utils/observability/utdTracker.test.ts
@@ -1,0 +1,498 @@
+// @vitest-environment jsdom
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { DecryptionFailureCode } from 'matrix-js-sdk/lib/crypto-api';
+
+// vi.mock factories are hoisted above imports, so referenced variables must
+// be prefixed with "mock" (Vitest's hoisting convention) -- see
+// https://vitest.dev/api/vi.html#vi-mock
+const mockAdd = vi.fn();
+const mockCreateCounter = vi.fn(() => ({ add: mockAdd }));
+const mockGetMeter = vi.fn(() => ({ createCounter: mockCreateCounter }));
+
+vi.mock('@opentelemetry/api', () => ({
+	metrics: { getMeter: mockGetMeter }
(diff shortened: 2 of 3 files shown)


Storypapst:
Reviewed: contract matches exactly (meter/counter names, category classification for all 12 DecryptionFailureCode values, no PII/identifying attributes ever attached). Good call building a fresh global-scope tracker on Room.timeline instead of reusing #412's single-room-scoped mechanism — reusing it would have missed every room not currently open in the UI. Correction on the crypto-init question you flagged: initRustCrypto() IS called (matrixClientService.ts:113, added 2026-07-10) — my earlier research finding was wrong, filed and closed as ORISO-Frontend#464. Good news for this PR: the tracker should see real organic data, not just synthetic tests. All 4 CI checks green. Merging.

PR #533, opened 2026-07-19
fix: allow consultants to disable 2FA

shazia-k:
## What changed

- Allow consultants with active two-factor authentication to disable it from the security settings dialog.
- Remove the development-toolbar-only exception that previously controlled the disable action after mandatory 2FA became active.
- Add regression coverage for an active consultant editing mandatory 2FA settings.

## Why

The security page showed the current second factor and allowed editing the authentication method, but hid the disable action for consultants after the mandatory-2FA date. This left consultants unable to turn off 2FA through the UI.

## Validation

- `npm run test:unit -- --run src/components/twoFactorAuth/TwoFactorAuth.test.tsx` — passed
- ESLint and Prettier checks for both changed files — passed
- `git diff --check` — passed
- Full repository gates were also attempted. They remain blocked by unrelated existing issues: missing TipTap/emoji packages, broken `localStorage` test setup, and existing SCSS lint violations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified two-factor authentication disable controls so they accurately reflect whether two-factor authentication is active.

* **Tests**
  * Added coverage verifying that active two-factor authentication opens the setup dialog with disabling enabled.

### DIFF
--- a/src/components/twoFactorAuth/TwoFactorAuth.test.tsx
+++ b/src/components/twoFactorAuth/TwoFactorAuth.test.tsx
@@ -1,0 +1,83 @@
+// @vitest-environment jsdom
+
+import * as React from 'react';
+import { render } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { UserDataContext } from '../../globalState';
+import { TwoFactorAuth } from './TwoFactorAuth';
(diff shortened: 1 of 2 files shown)


Storypapst (approved):
All good

Issue #541, opened 2026-07-19
Wire send-failure to the timeline (make the failed delivery state real)


Storypapst:
`Affected repos:` ORISO-Frontend · Parent: #537

## Why
#537 restored the "sending failed" delivery state (red cross) and the `MessageSendFailed` card, but nothing set it — a rejected send only called `apiPostError(...)` and never appeared in the conversation, so the state was a placeholder. Now that Matrix crypto is durable (ADR-004 amendment 2026-07-19) send failures are real and worth surfacing.

## What to build
- `messageSubmitInterfaceComponent`: add an `onSendError(message, ts)` callback, fired from the text-send `.catch` (the encryption/transfer failure path). The composer keeps the typed text so the user can resend.
- `SessionItemComponent`: hold failed sends in state, render a `MessageSendFailed` card at the bottom of the timeline for each, clear on the next successful send and on conversation change.
- Attachment sends keep their existing size/format error UI.

## Acceptance
- [ ] A rejected text send shows the "Sending message failed" card in the timeline (not only an error-API log).
- [ ] The composer text is retained on failure.
- [ ] Cards clear on the next successful send and when switching conversations.
- [ ] `tsc` + `eslint` green.

Storypapst:
Verified as delivered: PR #544 (merged to pre-dev) wires a real send failure to the timeline: `messageSubmitInterfaceComponent` now fires `onSendError` from the text-send catch path while keeping the typed text, and `SessionItemComponent` renders a `MessageSendFailed` card per failed send, clearing it on the next successful send or on conversation change; `tsc` and `eslint` verified clean.

Moving to **QA external Dev** for external review. cc @BjoernLudwig

Issue #537, opened 2026-07-19
Chat message redesign fixes: reactions UX, delivery status, avatar ring, measurements, send-failed state


Storypapst:
## Why

The chat message component drifted from the Figma design (App.Oriso, nodes 8498-32645, 336-12244/12225, 7539-29191/29194, 666-25242, 8609-31260) and several regressions were reported during design review with annotated screenshots:

1. The permanent quick-react smiley button overflowed the right viewport edge on own messages; reactions rendered inaccurately.
2. The delivered/read checkmarks ("zugestellt"/"gelesen") were missing entirely from the bubbles.
3. The technical username (`free_bee_frankie_821`) was displayed under own messages instead of a readable name.
4. The white avatar ring was too thin (2px) and the message-menu touch zone was below the required 32x32px.
5. Mobile measurements (bubble widths, 41px side insets, 48px date-row height) and desktop bubble widths (460px default, 770px for long texts) did not match the spec.
6. There was no "sending failed" message state (red cross in the time rail) and no Storybook component for it.
7. Chat-history card stack regression: 24px group corner radius missing in app view, avatar+name linkage overwritten, and the 6px gap must apply only around the ACTIVE card, not between all cards.
8. CI token-contract failure: styles consumed `--m3-*` tokens the OrisoScheme engine does not emit.

## What to build

- Move reactions into the bubble "Time + Emoji Rail" (incoming: chips left/time right; outgoing: time left/chips right) as white M3 assist chips with red own-count; quick-react emoji row lives at the top of the message action menu; Slack-style 500ms long-press opens the menu.
- Restore delivery status in the bubble time rail: single red checkmark = delivered, double = read, red cross = send failed.
- Humanize technical usernames in `messageNameUtils`.
- Avatar ring grows outward (60px frame, 6px ring) without touching kebab/bubble; 32x32 kebab touch zone.
- Enforce Figma measurements on mobile and desktop (bubble caps 460/770 on desktop, 41px insets, 48px mobile date row, 24px desktop date spacing).
- Add `MessageSendFailed` component + Storybook stories; reuse the Carimat bot avatar for system notifications.
- Session card stack: flush stacked group with 24px corners at group edges; only the active card gets full radius + 6px vertical gap; restore avatar/name linkage.
- Consume only engine-emitted `--m3-*` tokens.

## Acceptance

- [ ] Zero horizontal overflow on all message stories at 375px (touch) and 1280px (Playwright-measured).
[...]

PR #544, opened 2026-07-19
feat(chat): surface failed sends in the timeline

Storypapst:
Closes OpenResilienceInitiative/ORISO-Frontend#541 · Parent #537

Wires the "Sending message failed" card (Figma 7086-57415) to real send failures. Until now a rejected send only logged to the error API and nothing appeared in the conversation.

- `messageSubmitInterface`: new `onSendError(message, ts)`, fired from the text-send `.catch`; composer keeps the text so the user can resend.
- `SessionItemComponent`: `failedSends` state renders a `MessageSendFailed` card per failure at the bottom of the timeline; cleared on the next successful send and on conversation change.
- Attachment sends keep their existing size/format error UI.

Verified: `tsc --noEmit` 0 errors, `eslint` clean.

## Summary by CodeRabbit

* **New Features**
  * Added visible failure notifications when sending a message doesn’t succeed.
  * Preserved failed message text in the composer so it can be resubmitted.
  * Automatically clears failed-send notices after a successful send or when switching sessions.


shazia-k (approved):

Storypapst:
@coderabbitai full review

Issue #574, opened 2026-07-21
Preserve the original message before the send-failed system card


Storypapst:
Affected repos: ORISO-Frontend · Parent: #537

## Why

The outgoing-send error handler receives the original message text but stores only an id and timestamp. The timeline therefore replaces the failed user message with a generic system card. The card asks the sender to resend or copy a message that is not present as a failed timeline item.

## What to build

- Preserve the rejected outgoing message in local failed-send state.
- Render the original outgoing message as its own bubble with the red failed-delivery X.
- Render the explanatory `MessageSendFailed` system card immediately after that bubble.
- Add a primary-color text-only retry action that reuses the existing send pipeline, disables while in flight, and never retries automatically.
- Keep the failed pair local to the sender and remove it after a successful retry without sending the system card to the recipient.

## Acceptance

- [ ] A rejected text send shows the exact original text in an outgoing failed bubble.
- [ ] The failed bubble shows the red X and is followed immediately by the explanatory card.
- [ ] The recipient receives neither the rejected text nor the local system card.
- [ ] One click on `Retry send` performs exactly one normal send attempt; concurrent clicks and automatic loops are impossible.
- [ ] A successful retry produces one normal sent message and removes the stale failed pair.
- [ ] Focused component tests, full relevant test suite, lint/type-check, build, and real-browser desktop/mobile evidence are green.

Storypapst:
Verified as delivered: PR #575 (merged to pre-dev) preserves the original rejected message as its own failed bubble ahead of the `MessageSendFailed` system card, adds a single-attempt local-only retry action, and keeps the failure local to the sender; PR #588 (merged to pre-dev) closes the follow-up gap where a successful retry left a duplicate stale local echo, cancelling the rejected matrix-js-sdk echo by transaction id so exactly one message remains after retry.

Moving to **QA external Dev** for external review. cc @BjoernLudwig

PR #575, opened 2026-07-21
fix: preserve and retry failed messages

Storypapst:
## What changed

- Keep the original rejected text message visible beside a local failure explanation.
- Add a Primary text-only **Try again** action in the same status row as timestamp and delivery state.
- Replay the exact original room/thread, reply relation, mentions and aside routing through the existing send pipeline.
- Preserve a newer composer draft while retrying the failed message.
- Prevent duplicate failure cards, stale callbacks and parallel repeated retries.
- Provide the full explanation and action in English, German, informal German, French, Russian, Tigrinya and Turkish.

## Safety

This adds no endpoint or automatic retry loop. One manual click starts at most one normal send attempt, the action is disabled while that attempt is active, and existing server-side limits remain unchanged. The explanatory card is local UI state and is not sent to the recipient.

## Verification

- `npm run test:unit`: 190 files, 1,219 tests passed
- Production build passed (existing Autoprefixer and bundle-size warnings only)
- TypeScript, targeted ESLint, Stylelint and diff check passed
- Local browser proof passed on desktop and mobile for the renewed-device pre-transport failure path: original text retained, newer draft retained, one retry action, no duplicate card

The browser scenario intentionally proves failure recovery before transport; successful recipient delivery still requires the regular PreDev image after merge.

Closes OpenResilienceInitiative/ORISO-Frontend#574


shanzaeimran:
Reviewed #575 — the transport replay (prefixes, aside, reply, mentions) and single-flight guard are solid. 👍

Pushed a fix (98e80d0): success cleared all failure cards, so sending an unrelated message wiped other rejected messages and their retry. Now it only clears the card the retry actually resolves.

Three to weigh before merge:

Orphaned retry — switch threads while a retry is queued and no composer picks it up; retry buttons stay stuck disabled until a session switch.
Attachments dropped silently on retry (text-only) — fine if intended, just flag it on the card.
Rejected text shows twice (composer keeps it + card renders it).
New tests pass.

PR #579, opened 2026-07-21
feat(notifications): two-slot sound model + decoupled playback + sound dialog (#576)

Storypapst:
## Why

Notification sounds were the weakest surface: one hardcoded sound, a single on/off toggle, and the sound was chained to the OS popup (no sound with the tab focused or without browser-notification permission). Closes #576; implements the 2026-07-18 decision (sounds global, two slots) and Frank's UX: a **popup dialog from a menu entry**, reusing existing icons.

## What

- **Decoupled playback** — `playNotificationSound(family, isMention)` fires from `NotificationsProvider` on a genuinely new unread event, independent of focus/permission. Routed through the single `isNotificationSuppressed` gate (DND, per-conversation level, mute, family-off) and rate-limited (1 sound / 2 s). Volume 0.5.
- **Two global slots** — `sounds { message, mention }` with `SoundId` (`none|chime|ding|soft|default`; mention `default` inherits messages). Parse/merge/default like `dndUntil`, cross-device via account data.
- **Calls sound** — the `calls` family uses the existing `incomingCall.mp3` (Slack-style separate call sound).
- **Sound dialog** — `SoundSettingsDialog` (OrisoDialog, hero `audio-on` icon) opened from a menu entry that replaces the old on/off switch; per-slot dropdown + ▶ preview, message slot hides `default`. i18n de/de@informal/en.

Sound assets are placeholders (only two exist today) — the id model + UI already carry the distinct choices; swap the asset map when a curated library lands.

## Verification

TDD: two-slot model (4), `soundPlayback` helpers (8), dialog render (5). Full suite **1228 green**, tsc + eslint clean; Storybook dialog body verified (two slots, preview enabled for chime, disabled for the inherited-silent mention).

Design: `0 - Docs/PROPOSAL-notification-sounds.md`. Composes with the shipped DND gate (#571) and per-conversation level.

Closes OpenResilienceInitiative/ORISO-Frontend#576

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a dialog-based notification sound setup for new messages and `@` mentions, including tone selection and audio preview.
  * Introduced per-area notification sound configuration with volume controls and optional “send by email” settings.
  * Notification sounds now play for newly received unread events according to your selected message/mention tones.
* **Bug Fixes**
  * Prevented repeated sound announcements on initial load and during subsequent feed refreshes.
  * Added throttling to reduce excessive playback.
* **Documentation**
[...]

Issue #576, opened 2026-07-21
Notification sounds: two global slots + popup dialog from a menu entry, decoupled playback


Storypapst:
**Affected repos:** ORISO-Frontend

## Why

Notification sounds today are the weakest part of the notification system: **one hardcoded sound, a single on/off toggle, and the sound is chained to the browser (OS) notification** — so with the tab focused or without OS-notification permission, nothing plays. The grill decision (2026-07-18) was "sounds global, two slots (messages + mentions), each incl. 'no sound'"; that is decided but not built.

Product decision (Frank, 2026-07-18): configure sounds via a **menu entry that opens a popup dialog** (simplest, cleanest UX — same popup pattern as the per-conversation menu in #420), reusing existing repo icons. The exact sound *choices* can be tuned later; build the foundation now.

Full analysis: `0 - Docs/PROPOSAL-notification-sounds.md`.

## What to build

**1. Decouple sound from the OS popup (the key correctness fix)**
- Extract playback into `playNotificationSound(family, isMention)` called when a new feed event arrives (from `NotificationsProvider`), independent of window focus / browser-notification permission.
- Keep routing through the single gate `isNotificationSuppressed(settings, device, family, now)` so DND, the per-conversation level, and mute all still suppress sound.

**2. Two global sound slots in the cross-device settings model**
- Replace `sounds: { enabled }` with `sounds: { message: SoundId; mention: SoundId }`, `SoundId = 'none' | 'chime' | 'ding' | 'soft' | 'default'` (mention `'default'` = same as message). Parse/merge/default like `dndUntil`; syncs via account data.

**3. Popup dialog reached from a menu entry**
- A `SoundSettingsDialog` (Modal/popup) with, per slot, a dropdown (Kein Ton / Chime / Ding / Soft) and a **▶ preview** button (`resolveSoundAsset` → play).
- Reachable from a menu item (icon: reuse `notification_bell.svg` / `audio-on.svg`; "none"/muted state: `audio-off.svg` / `bell-off.svg`). Follows M3 dialog rules (hero icon, mobile stacked footer).

**4. Calls slot for free**
- Route the `calls` family to the existing `incomingCall.mp3` (Slack-style separate call sound) — no new UI.

**5. Nice-to-haves**
- Rate-limit / debounce (max ~1 sound / 2 s) so active groups don't rattle.
- Fixed volume ~0.5; swallow autoplay-block errors (already done).

[...]

PR #571, opened 2026-07-21
feat(notifications): global Do-Not-Disturb control + announcement suppression

Storypapst:
## Why

Frank's follow-up: a global "Nicht stören" (Slack-style) that silences all announcements for a chosen time and auto-reverts. Pairs with the UserService store + email gate (ORISO-UserService#533). The persisted activity feed keeps filling — only the announcement is silenced.

## What

- **One gate covers all three surfaces:** `isNotificationSuppressed` now consults `dndUntil` first, and it already fronts toast + notification sound + browser push in `sendNotification`, so DND silences all three at once. New pure `isDoNotDisturbActive(dndUntil, now)`.
- `dndUntil` added to the cross-device notification settings model (parse/merge/default) → syncs across devices via the existing account-data store.
- `computeDndUntil(option, now)` maps 1h / 8h / until-tomorrow → ISO. `DoNotDisturbControl` writes **both** the settings store (announcements) and the UserService endpoint (email gate) via `apiDoNotDisturb`.
- Wired into the profile Notification Settings panel; i18n de / de@informal / en.

## Verification

TDD: `isDoNotDisturbActive` + gate-honours-DND (5), `computeDndUntil` (4), `DoNotDisturbControlView` render (3). Full suite **1206 green**, tsc + eslint clean; Storybook active/off states verified (screenshot: "Nicht stören · Aktiv bis … · Aus / 1 Stunde / 8 Stunden / Bis morgen früh").

Part of the notification follow-ups (Team-Besprechung epic OpenResilienceInitiative/ORISO-Frontend#512).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a global Do Not Disturb setting to notification preferences.
  * Users can pause notifications for 1 hour, 8 hours, until tomorrow morning, or indefinitely.
  * Displays when Do Not Disturb will end and synchronizes the setting across devices.
  * Notifications are suppressed while Do Not Disturb is active.

* **Documentation**
  * Added English and German translations for the new setting.

* **Tests**
  * Added coverage for duration selection, expiration handling, display states, and notification suppression.

PR #586, opened 2026-07-22
feat(notifications): config-driven playback, Safari hardening, arrow icons, panel spacing

Storypapst:
Follow-up to #579 for issue #576 — makes the notification-config dialog actually drive playback, and hardens the pipeline for Safari.

### Config dialog now drives real playback
- New event → (area × kind) mapping shared by UI and playback (`notificationConfig.ts`): family `requests`→Anfrage, `appointments`→Termine, everything conversation-shaped→Gespräch; kind = `mention` > arrival events (`request.new`, `team.discussion.new`, `waiting_room.client.joined`) → `new` > `standard`.
- `playNotificationSound` plays the configured tone **at the configured per-row volume**; `calls` keep the dedicated ring. Previously the dialog persisted config that playback never read.

### Safari compatibility (verified against WebKit docs)
- `requestNotificationPermissionSafe()`: resolves in both the promise and legacy-callback forms, never throws.
- `useNotificationPermission`: requests on the user's **first pointer/keyboard gesture** instead of a 2s timer — Safari silently ignores non-gesture prompts, Chromium down-ranks them.
- Audio unlock: one shared `<audio>` element is primed with a silent WAV on the first gesture; event sounds reuse it so Safari's autoplay policy (NotAllowedError without gesture) doesn't swallow them.

### UI polish (Frank's review)
- Volume arrows use the keyboard_arrow up/down icons from the Icons Master File — bare, hover → black.
- Settings panel spacing: styled "Benachrichtigungen einstellen" entry button, NotificationDenied box margins, section-title gap, divider rhythm. Styles live next to the component so Storybook shows the real thing.

## Tests
+15 unit tests (1260 total green): area/kind mapping, config-driven playback incl. volume/mute/calls/none, audio unlock priming, `sendNotification` permission gate (no OS popup without browser permission + opt-in), requestPermission promise & callback forms.

## Out of scope
Background Web Push (service worker + VAPID) — the app only notifies while a tab is open; noted in #576 as a possible follow-up (Safari 16.4+/macOS 13+ supports standard VAPID push, iOS only as installed home-screen app).

## Summary by CodeRabbit

- **New Features**
  - Notification configuration now supports per-area/per-kind sound + volume, plus banner modes (off/temporary/persistent), including time-critical “Wecker” options.
  - Added a new styled checkbox control and expanded sound choices (including a ring tone).
  - Added a global notification-status toggle in the navigation with a sidebar placement option.

- **Bug Fixes**
  - Browser notification permission is requested only after user interaction and works across modern and older browser behaviors.
  - Improved Safari audio playback reliability for notification sounds and better handling when notifications are unsupported/denied.

---

## Second commit: harmonised notification model (Frank's design review 2026-07-22)

- **Inline tabs in the settings page** (Anfrage / Gespräch / **Zeitkritisch**) replace the flat per-family toggle list; changes save immediately. The dialog wrapper stays as quick access from the conversation ⋮-menu (new SessionMenu entry) so users are not thrown out of their room.
- **Three channels per row**: 🔔 banner (browser popup) · 🔊 sound + volume · ✉️ email — full congruence, `sendNotification` now consults the banner channel of the event's config row.
- **Gespräch gains an Übergabe row** (handover.* / case.handover.*); **Zeitkritisch** hosts *Eingehender Anruf* (default: the dedicated ring, now user-configurable) and *Termin*.
- **Drafts never notify** on any channel (feed/timeline only — display-filter follow-up in #592); **System** stays a single global switch outside the tabs.
- **Wecker-Modus feature-signal dummy** (#590) with 👍/👎 in the Zeitkritisch tab; backend vote API is OpenResilienceInitiative/ORISO-UserService#544.
- Tests updated + extended; full suite 1264 green; Storybook (panel + dialog) verified.

---

## Third commit: 3-state banners, rail status button, M3 checkboxes

- **Banner mode per row**: off / temporary / **persistent** — persistent sets `requireInteraction` so the banner stays until dismissed. Chromium honours it; Firefox ships it disabled, Safari never supports it (macOS system settings decide) — both fall back to a temporary banner. Sources: [caniuse](https://caniuse.com/mdn-api_notification_requireinteraction), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Notification/requireInteraction).
- **Global status button in the navigation rail** (opt-in from the settings page, mirroring the Live-Chat rail toggle): flips the account-wide `globalMute` — the switch that overrides everything. Bell / bell-off states with a11y labels.
- **M3Checkbox** component (18dp box, 2dp radius, primary fill, 40dp state layer) replaces the raw accent-color checkboxes flagged in design review.
- Legacy boolean banner values parse tolerantly; 1266 tests green.


Storypapst:
@coderabbitai All 16 review findings addressed in 1cc9c52:

- **Functional**: SessionMenu entry → real `<button>` (keyboard); banner select :focus-visible ring restored; `useNotificationPermission` reports 'denied' when unsupported; status-toggle hook consumes `CustomEvent.detail` (restricted-storage Safari); suppression gate consumes `NEVER_NOTIFY_FAMILIES`.
- **Maintainability**: shared `onFirstUserGesture` util (permission ask + audio unlock); shared `previewNotificationSound` (was copy-pasted ×3); `__resetSharedAudioForTests` + afterEach reset (no test order-dependency); registry import instead of barrel; nav icon rides `--oriso-nav-foreground`/`--oriso-nav-active-foreground`; M3Checkbox token pass (note: `--m3-state-hover` is NOT emitted by the OrisoScheme engine — the m3Sweep contract test enforces this — so the state layer keeps a literal); decorative icons `aria-hidden`; dead `soundEntry`/`__email` CSS removed.
- **Tests**: +7 (throw/reject permission paths, status-toggle incl. storage-fail, M3Checkbox, all-tabs assertion). Suite: 1273 green.

The full-token "heavy lift" for the notification controls stays open for Frank's announced design overhaul.

Storypapst:
@coderabbitai review

Branch is now conflict-free with pre-dev (merge commit a239938): the global-DND package (#571) and the selectEventToAnnounce refactor are integrated — DND-until joins the suppression gate ahead of the harmonised per-family rules, the provider passes eventType through to config-driven playback, and the panel gains DoNotDisturbControl above the inline tabs. Full suite 1310 green, eslint/stylelint clean. Please re-review so the stale CHANGES_REQUESTED clears.

shanzaeimran (approved):

PR #776, opened 2026-07-25
fix(notifications): call-ring reliability, banner routing, remove dead sound model

Storypapst:
Hardening pass over the notification package now that #586 is on pre-dev. Three real defects, one cleanup.

## 1. A message tone could swallow the incoming-call ring 🔴
One global 2s throttle covered **all** families. A message arriving 100ms before a call meant the ring never played — the most time-critical sound we have was the easiest to lose.

**Fix:** one throttle bucket per family. Bursts inside a family are still collapsed; a call is never blocked by chat traffic.

## 2. …and the ring could be cut off mid-play 🔴
The Safari-primed shared `<audio>` element was reused unconditionally: assigning `.src` while the ring was playing killed it.

**Fix:** reuse the shared element only when it is idle; while it is busy a fresh element is used. In Safari the worst case is now "the ring keeps playing", which is the right failure mode.

## 3. The enquiry banner was routed to the wrong config row 🟠
Neither `sendNotification` call site passed `eventType`/`family`, so every OS banner resolved to **Gespräch → Standard**. Turning banners off for "Neue Anfrage" did nothing; turning them off for Gespräch/Standard silently killed enquiry banners too.

**Fix:** both call sites pass `family` + `eventType`.

## 4. Cleanup: the superseded two-slot sound model
No consumers since the config tabs took over — removed `SoundSettingsDialog` (+ test + story), `settings.sounds`, `resolveEventSound`, `assetForEvent`, `asSoundId` and the `profile.notifications.sounds` i18n block.

## Known gap (NOT fixed here, deliberately)
Only **two** events can produce an OS banner at all (`message.new`, `request.new`). The banner column is therefore still decorative for the Erwähnung / Übergabe / Anruf / Termin rows. The clean fix is to consolidate banner + sound into the single announcement path in `NotificationsProvider` — it already resolves family, eventType and mention — and retire the two legacy call sites. That touches the live message-notification path, so I would rather land it **after** the long-form E2E (OpenResilienceInitiative/ORISO-E2E#20) can prove it. Happy to ticket it.

## Verification
1302 unit tests green (3 new: ring-not-swallowed, still-throttles-within-a-family, no-cutoff), tsc/eslint/stylelint clean. Drive-by: two stylelint comment-spacing violations inherited from earlier merges.

## Summary by CodeRabbit

- **New Features**
  - Notifications for new messages and requests now use improved categorization and routing.

- **Enhancements**
  - Notification sounds are managed more reliably, with separate throttling for different notification types.
  - Sound playback now avoids interrupting an already-playing notification.

- **Changes**
  - Removed the notification sound selection settings and related translation text from profile settings.


Storypapst:
The "known gap" section is now a concrete plan instead of a note: **EPIC #778** with three sub-issues — #779 (consolidate banner + sound into one announcement path), #780 (calls leave a durable message in the room timeline), and OpenResilienceInitiative/ORISO-UserService#730 (emit `call.*`/`appointment.*` notifications). All on the Roadmap board.

shanzaeimran (approved):

Issue #24, opened 2025-11-16
[Feature Request]: Image compression and thumbnail view with click-to-expand


nikunjdecyb:
### Target Service/Component

Frontend (ORISO-Frontend)

### Priority

P2-Medium

### Problem Statement

Large images displayed at full size - overwhelming, slow to load, consumes excessive bandwidth. No thumbnail preview with click-to-expand functionality.

### Proposed Solution

- Compress images (client-side or server-side)
- Display thumbnails in chat (200-300px max)
- Click thumbnail to open full-size in modal/lightbox
- ESC key and close button to return to chat

### Technical Considerations

- Client-side: browser-image-compression library
- Server-side: Image processing libraries
- Thumbnail generation (Canvas API or server-side)
- Modal/lightbox component
- Image storage strategy

### Related Documentation/Links

https://github.com/OpenResilienceInitiative/ORISO-Frontend/issues/20

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't been requested before
- [x] I have provided a clear problem statement and proposed solution
- [x] This feature aligns with ORISO's mission of accessible mental health support

kennstenicht:
@nikunjdecyb Could you please check if this issue is done? If so, please close it.

Storypapst:
Closing as completed — backlog re-audit found this already resolved: PR #547 merged — includes image preview/thumbnail functionality.

PR #783, opened 2026-07-25
Confine call rooms to session members, add widget-mode embedding

Storypapst:
## 1. The call room was joinable by anyone who knew its id

`CallManager.createElementCallRoom` created the per-call room with `preset: 'public_chat'`, which sets `join_rule: public`. Room ids are not secrets — they travel in the call invite event, in logs and in support tickets — so **anyone who learned one could join a counselling call**.

It is now created private with a `restricted` join rule allowing exactly the members of the parent session room, plus `history_visibility: joined` and guest access off.

A homeserver that rejects restricted joins (room version < 9) falls back to invite-only with explicit invites — **never back to a public room**.

## 2. Element Call can now be embedded as a Matrix widget

In widget (matryoshka) mode the iframe receives **no access token** and registers **no device of its own**. It asks this client to send and read events over `postMessage`, and the host client — which already has crypto set up — does the Matrix I/O.

`OrisoWidgetDriver` is purpose-built rather than general:

- **Bound to the single call room.** Any request naming another room is refused, so a call widget cannot read the counselling conversation.
- **Explicit capability allow-list.** The iframe cannot send `m.room.message` or change join rules as the user. Unknown capabilities are denied rather than granted.
- **Encrypted to-device sends go through the host crypto stack** and throw if it is missing, instead of silently downgrading the media keys to plaintext.

### Off by default

Behind `REACT_APP_ELEMENT_CALL_WIDGET_MODE`. The widget path has **not** been exercised against a deployed Element Call yet; the existing device-bound path stays the default until it has been. Flipping the flag on Pre-Dev is the next step.

## Testing

21 new cases:

- capability allow-list, including state-key smuggling (`…power_levels#org.matrix.msc3401.call.member` must not be granted)
- driver room confinement, on both read and write
- encrypted vs plain to-device routing, and the refusal when crypto is absent
- the call room's join rule, its encryption/history/guest state, and the invite fallback

## Summary by CodeRabbit

* **New Features**
  * Added an option to embed Element Call as a Matrix widget, including a runtime toggle, iframe setup, and widget-to-host event/to-device communication.
  * Introduced widget-mode capability allow-lists and a widget-mode driver that forwards calls to the host with filtering.
  * Call-room creation now prefers restricted join rules for counselling-session members, with encryption enabled and guest access disabled.
* **Bug Fixes**
  * Improved widget-mode handling for missing/invalid widget session data and tightened enforcement of room confinement and capability validation.
* **Tests**
  * Added/expanded unit tests for widget capabilities, widget driver behavior (events, to-device encryption, OpenID), and call-room creation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
 		"lottie-react": "^2.4.1",
 		"markdown-draft-js": "^2.4.0",
 		"matrix-js-sdk": "^38.4.0",
+		"matrix-widget-api": "^1.13.1",
 		"mini-css-extract-plugin": "^2.7.7",
 		"prompts": "^2.4.2",
 		"qrcode": "^1.5.3",
--- a/src/components/call/GroupCallWidget.tsx
+++ b/src/components/call/GroupCallWidget.tsx
@@ -8,8 +8,10 @@ import React, { useEffect, useState, useRef } from 'react';
 import { callManager, CallData } from '../../services/CallManager';
 import {
 	getElementCallBaseUrl,
-	getMatrixHomeserverUrl
+	getMatrixHomeserverUrl,
+	isElementCallWidgetModeEnabled
 } from '../../resources/scripts/runtimeConfig';
+import { useElementCallWidget } from './widget/useElementCallWidget';
 import { useMatrixClient } from '../../globalState/context/MatrixClientContext';
 import { getElementCallAccessToken } from '../sessionCookie/getMatrixAccessToken';
 import {
@@ -33,6 +35,21 @@ export const GroupCallWidget: React.FC = () => {
 	const [callState, setCallState] = useState<string | null>(null);
 	const [elementCallUrl, setElementCallUrl] = useState<string>('');
 	const [isDismissed, setIsDismissed] = useState(false);
+
+	// Widget mode keeps the Matrix session in this app: the call iframe gets no
+	// access token and registers no device of its own, and asks us to send and
+	// read events for it instead. Off by default until validated on Pre-Dev.
+	const widgetModeEnabled = isElementCallWidgetModeEnabled();
+	const callRoomId = callData
+		? ((callData as any).elementCallRoomId ?? callData.roomId)
+		: null;
+	const widget = useElementCallWidget(
+		widgetModeEnabled ? (matrixClientService?.getClient() ?? null) : null,
+		{
+			roomId: widgetModeEnabled ? callRoomId : null,
+			isVideo: callData?.isVideo ?? true
+		}
+	);
 
 	// Dragging / resize state
 	const [isDragging, setIsDragging] = useState(false);
@@ -200,6 +217,22 @@ export const GroupCallWidget: React.FC = () => {
 
 	const setupElementCall = async () => {
 		if (!callData || setupInProgressRef.current) return;
+
+		// In widget mode the URL is derived declaratively by the hook — there is
+		// no login step to perform here, because there is no separate session.
+		if (widgetModeEnabled) {
+			if (widget.error) {
+				alert(`Failed to start call: ${widget.error.message}`);
+				callManager.endCall();
+				return;
+			}
+			if (widget.url) {
+				setupInProgressRef.current = true;
+				setElementCallUrl(widget.url);
+			}
+			return;
+		}
+
 		setupInProgressRef.current = true;
 
 		try {
@@ -627,7 +660,17 @@ export const GroupCallWidget: React.FC = () => {
 									: '⤢'}
 						</button>
 						<iframe
-							ref={iframeRef}
+							ref={(node) => {
+								// The ref is shared: the drag/close logic needs the
+								// element, and widget mode additionally opens the
+								// postMessage channel against it.
+								(
+									iframeRef as React.MutableRefObject<HTMLIFrameElement | null>
+								).current = node;
+								if (widgetModeEnabled) {
+									widget.attachIframe(node);
+								}
+							}}
 							src={elementCallUrl}
 							className="element-call-iframe"
 							allow="camera; microphone; display-capture; autoplay; fullscreen"
--- a/src/components/call/GroupCallWidget.tsx
+++ b/src/components/call/GroupCallWidget.tsx
@@ -4,7 +4,7 @@
  * https://github.com/element-hq/element-call
  */
 
-import React, { useEffect, useState, useRef } from 'react';
+import React, { useCallback, useEffect, useState, useRef } from 'react';
 import { callManager, CallData } from '../../services/CallManager';
 import {
 	getElementCallBaseUrl,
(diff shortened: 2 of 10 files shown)


Storypapst:
### Added: OpenID bridging (a20993e)

The driver was missing `askOpenID`, which turns out to be load-bearing rather than optional: Element Call needs an OpenID token to authenticate against the LiveKit JWT service. `WidgetDriver`'s default implementation answers `Blocked`, so **widget mode would have failed to connect a call at all** — the flag could never have been turned on.

`askOpenID` now bridges to `client.getOpenIdToken()` and validates the response before handing it on (non-empty `access_token` / `token_type` / `matrix_server_name`, finite positive `expires_in`). Anything malformed, rejected or thrown answers `Blocked`, and the rejection reason is never passed to the widget.

One deliberate trade-off worth flagging for review: the failure paths are **silent** — the tests assert nothing is written to the console. That keeps a rejection from leaking anywhere, but it also means a call that fails to connect for OpenID reasons leaves no trace to debug. If we hit that on Pre-Dev, a single non-sensitive log line is the fix.

Test count for this branch is now **34** (21 + 13 OpenID cases), all passing.

Storypapst:
## Deployed to Pre-Dev and verified by a real call — **green**

This branch is running on Pre-Dev, and the encryption gate (ORISO-E2E#32) now **passes** against it. Two consecutive runs, exit 0.

A real 1:1 counselling call was started through the product path. The call room it created:

```json
{
  "joinRule": "restricted",
  "allow": [
    { "room_id": "!ELBEJIcvfpwjwKenkh:91.99.183.160", "type": "m.room_membership" }
  ],
  "encryptionAlgorithm": "m.megolm.v1.aes-sha2",
  "historyVisibility": "joined"
}
```

- **`restricted`** with the parent session room as the allow rule — only people already in that counselling session can join the call.
- **Encrypted**, which is what makes Element Call select per-participant media encryption rather than falling back to plaintext.
- **History closed** to non-members.
- Element Call did **not** raise `E2EEUnavailableError`, so the crypto stack was present and the call was not silently downgraded.

### The before/after is the strongest part

Immediately before this deployment, the **same test on the same environment failed**:

```
Error: call room must not be publicly joinable
Expected: not "public"
```

Pre-Dev was creating `join_rule: public` call rooms — anyone holding a room id could have joined a counselling call. That is a validated negative control: the gate caught the real vulnerability, and this branch is what closes it.

### Deployment

Built natively on the Pre-Dev host (`docker build`, `k3s ctr images import`, `kubectl set image`). Pre-Dev now runs an **unmerged branch build** of the frontend. Rollback:

```bash
ssh root@46.224.170.69 "kubectl set image deploy/oriso-platform-frontend -n caritas frontend=ghcr.io/openresilienceinitiative/oriso-frontend@sha256:d3a5a771daf57ef59ed3ab610ad643d63993e47364612d03ee43be4ffbf46241"
```

Element Call on Pre-Dev is likewise running ORISO-ElementCall#24 (rollback command in that PR).

Storypapst:
## Review findings fixed — including one that defeated this PR's own fix

Five of CodeRabbit's findings were real. One deserves to be called out because it made the headline change worthless:

### The restricted join rule was defeatable from inside the call

I set `join_rule: restricted` correctly — and then left `state_default: 0`, which let **any joined participant overwrite `m.room.join_rules` with `public`**. A lock fitted to a door anyone could rehang. State changes now require the creator's level, with `join_rules`, `guest_access`, `history_visibility`, `canonical_alias`, `server_acl`, `encryption`, `name` and `power_levels` all pinned at 100. The one exception stays at 0: `org.matrix.msc3401.call.member`, because Element Call must publish membership from every participant.

### The other four

| Finding | Fix |
|---|---|
| **CWE-732** — unscoped `send.state_event:m.room.member` was granted; the widget could change room membership as the user | Split into a narrow send list (call membership only) and a wider receive list |
| **CWE-319** — `sendToDevice` honoured the widget's `encrypted` flag for `m.call.encryption_keys` too, and the iframe controls that flag | Key-bearing types are refused outright when plaintext is requested |
| **CWE-668** — `feedToDevice` forwarded *every* to-device event (device verification, room keys) into the iframe, which had only asked for call keys | Filtered to the allow-list |
| **CWE-532** — a failed invite logged the Matrix user id and raw error from a production browser | Replaced with a count; in a counselling product a Matrix id identifies a person |

### Re-verified on Pre-Dev after the fixes

Rebuilt, redeployed, and ORISO-E2E#32 re-run — **still passing**, real call, real room:

```json
{
  "joinRule": "restricted",
  "allow": [{ "room_id": "!ELBEJIcvfpwjwKenkh:91.99.183.160", "type": "m.room_membership" }],
  "encryptionAlgorithm": "m.megolm.v1.aes-sha2",
  "historyVisibility": "joined"
}
```

New unit guards: media keys cannot be sent unencrypted, and a call participant cannot reopen the room — the latter asserts the power levels directly, since that is exactly where the first fix was quietly defeated. 38 passing in the call package.

### Still open from the review

Non-security items I have not done: memoizing the iframe ref callback, `widgetIdForRoom` collision hardening, and coverage for the delayed-event and read-timeline driver methods. Happy to take those in this PR if you want them before merge.

Storypapst:
@coderabbitai review

All 11 findings are addressed. Mapping, so the re-review has something to check against:

**Security**

| Finding | Fix |
|---|---|
| CWE-732 — unscoped `send.state_event:m.room.member` granted | Split into `ALLOWED_SEND_STATE_EVENT_TYPES` (call membership only) and `ALLOWED_RECEIVE_STATE_EVENT_TYPES` |
| CWE-862 — `state_default: 0` let any participant rewrite `join_rules` back to `public` | `state_default: 100`, plus explicit 100 on `join_rules`, `guest_access`, `history_visibility`, `canonical_alias`, `server_acl`, `encryption`, `name`, `power_levels`; only `org.matrix.msc3401.call.member` stays at 0 |
| CWE-319 — `sendToDevice` honoured the widget's `encrypted` flag for key-bearing types | Key-bearing types refuse plaintext outright |
| CWE-668 — `feedToDevice` forwarded all host to-device traffic into the iframe | Filtered to `ALLOWED_TO_DEVICE_EVENT_TYPES` |
| CWE-532 — failed invite logged the Matrix user id and raw error | Replaced with a count |

**Non-security**

| Finding | Fix |
|---|---|
| Inline ref callback rebuilt the widget channel on every render | Memoised with `useCallback`, declared after the values it closes over |
| `widgetIdForRoom` collided across distinct room ids | Encoded instead of stripped |
| Delayed-event and read-timeline methods untested | Added: delayed state event, delayed-path room confinement, `updateDelayedEvent`, timeline `limit` and `since` semantics |

**The `state_default` one deserves singling out**: it defeated this PR's own headline change. The restricted join rule was set correctly and then left rewritable from inside the call. Good catch — the unit guard now asserts the power levels directly, since that is where the fix was quietly undone.

44 tests passing in the call package, lint clean. Rebuilt and redeployed to Pre-Dev after the fixes; the encryption gate (ORISO-E2E#32) still passes against a real call: `joinRule: restricted`, `m.megolm.v1.aes-sha2`, `historyVisibility: joined`.

Storypapst:
@shanzaeimran @Shirloin — seven PRs landed together and they depend on each other, so here is a suggested order and what each one actually needs from a reviewer.

**Start here (security-critical):**

1. **ORISO-Frontend#783** (this one) — call room access + widget driver. The thing worth your attention: `state_default` in `CallManager.createElementCallRoom`. An earlier version of this PR set `join_rule: restricted` and then left `state_default: 0`, which let any joined participant rewrite the join rule back to `public` from inside the call. If that reasoning is wrong anywhere, the whole PR is wrong.
2. **ORISO-ElementCall#24** — media E2EE re-enabled. The claim to attack: that Element Call now has its own Matrix device (`getElementCallAccessToken`) and therefore *can* do per-participant encryption. If that is false, media E2EE should stay off and the fail-closed guard in `GroupCallView` is the wrong shape.

**Then:**

3. **ORISO-E2E#32** — the gate that proves 1 and 2. It has been shown to fail on the known-bad state (`join_rule: public`) and pass on the fixed one, both against Pre-Dev.
4. **ORISO-UserService#751** — makes the `RCToken` header optional. Check the dead-code trace: `UserSessionControllerDelegate:295` substitutes `"dummy-rc-token"`, and `RocketChatRoomInformationProvider` never reads the token when `rocket-chat.enabled=false`.
5. **ORISO-Frontend#784** — Rocket.Chat purge. Note it also fixes six components that my own route-param rename had silently broken.

**Independent, low risk:**

6. **ORISO-ElementCall#27** — lockfile-only bump closing all 16 runtime Dependabot alerts.
7. **ORISO-ElementCall#26** — `-s ours` merge recording upstream ancestry. Zero file changes; the only question is whether `1006d06b` is the right ancestor.

**Heads-up on Pre-Dev:** it currently runs unmerged builds of #783 and #24, because that is how the encryption was verified. That state is *safer* than before (the previous one had public call rooms and plaintext media), so there is no rush — but it should end up merged rather than lingering. Rollback commands are in each PR if you prefer.

shanzaeimran (approved):

Issue #748, opened 2026-07-22
fix: correct Keycloak logout and public-route 401 handling


Storypapst:
Backfilled during the GitHub hygiene sweep — PR #95 (author @shazia-k) shipped this work on **2026-06-07** but had no linked issue.

Maps to the historical NocoDB Work Package **Refactoring** — this work predates/extends that roadmap tracking, now reconciled into GitHub.

Storypapst:
Completed 2026-06-07 via PR #95. Backfilled + closed as part of the GitHub hygiene sweep.

PR #784, opened 2026-07-25
Remove Rocket.Chat from the frontend

Storypapst:
## Why

Rocket.Chat has been switched off for a while (`rocket-chat.enabled=false`, ADR-004), but its name and its dead code paths were still everywhere in the frontend. **"Rocket.Chat" no longer appears anywhere in hand-written frontend source.**

No behaviour changes.

## What was removed

| | |
|---|---|
| **13 `.backup` files** | Left over from the Matrix migration, ~4000 lines. They were *tracked*, so they polluted every search. |
| **Dead branches in `IncomingVideoCall`** | Both handlers forked on `isMatrixRoom(rcGroupId)` with a Rocket.Chat fallback. Every room id is a Matrix room id now — the backend writes the Matrix room into that field (`CreateEnquiryMessageFacade:220`) — so the fallback was unreachable. |
| **`apiRejectVideoCall`** | Its only caller was that dead branch. |
| **`NonPlainRoutesWrapper`** | Restored Rocket.Chat E2EE keys; had since become a pass-through returning its own children. Inlined into `Routing.tsx`. |

Plus: the `:rcGroupId` route placeholder is now `:groupId` (an internal name, not part of any URL a user sees), and comments that described *current* behaviour in Rocket.Chat terms — e.g. `// For RocketChat sessions, use the standard check` — were rewritten to say what the code actually does.

## What is deliberately NOT touched

`rcGroupId`, `rcUserId`, `rcToken` and `rcValidation` **where they are wire fields**: request bodies, the STOMP video-call payload, and the `RCToken` / `RCUserId` headers that UserService still requires on archive endpoints.

Renaming those is a coordinated change with UserService — doing it here would break the API. It gets its own issue.

## Testing

Routing and selection suites pass (47 tests).

The failures in a full run are `src/utils/observability/*` and `src/components/productTour/*`, which fail on missing `@opentelemetry` / `react-joyride` modules in the local install. **Neither is touched by this change** — the diff and the failing files do not intersect.


Storypapst:
### Added: `rcValidation` → `sendChatUserHeaders` (5a2b936)

I went back through the boundary I had stopped at and found that one of the four names was never a wire field at all.

`rcValidation` is a **`fetchData` prop**. It decides whether to attach the `RCToken` / `RCUserId` headers and never leaves the browser — so it could be renamed freely all along. 28 call sites, now `sendChatUserHeaders`.

While tracing it I also established what the two headers actually do, which the old comment guessed at:

| Header | Status |
|---|---|
| `RCUserId` | **live** — becomes `RocketChatCredentials.rocketChatUserId` in UserService and drives the Matrix room lookup. Must stay. |
| `RCToken` | **dead** — threaded through five layers and discarded. `UserSessionControllerDelegate` even substitutes the literal `"dummy-rc-token"` when it is absent. |

The frontend was sending `MATRIX_MIGRATION_DUMMY_RC_TOKEN` to satisfy a check that no longer exists — both sides feeding each other a placeholder.

Four endpoints still returned 400 without it, so the token cannot simply be dropped here. **ORISO-UserService#751** makes the header optional; once that is deployed, removing `RCToken` from this file is a two-line follow-up.

The comment above the header block now says this instead of guessing.

Storypapst:
## I broke six components with this PR, and a test covered for it

Renaming the route placeholder `:rcGroupId` → `:groupId` in `RouterConfig` was only half the change. Six components still read the old name:

```ts
useParams<{ rcGroupId: string }>()
```

`AskerInfo`, `GroupChatInfo`, `CreateConversationView`, `SessionsList`, `AcceptAssign` and `SessionView` would every one have received `undefined` for the group id. Session view, session list, group chat info and the group-chat edit mode: **silently broken, no error anywhere.**

And `CreateConversationView.editmode.test.tsx` mocked `params: { rcGroupId: 'rc-1' }` — so it kept passing against a param name the router no longer produces. That is how this reaches Pre-Dev: a test that agrees with the old world.

All six readers and the mock now use `groupId`. **264 tests pass** across the affected component trees.

### Worth carrying into ORISO-UserService#748

The remaining `rcGroupId` / `rcUserId` rename is the same shape of change, one layer down. This PR is the evidence that a rename is *not* a search-and-replace on the declaration site:

- every reader has to move with it, and they are not co-located with the declaration
- any test that mocks the old shape will cover for the mistake rather than catch it

For #748 that means the additive-alias step is not optional politeness — it is what makes the readers migratable one at a time instead of all-or-nothing.

### Also in this PR since the last review

`rcValidation` → `sendChatUserHeaders` (28 call sites). It was never a wire field — a `fetchData` prop that never leaves the browser — so it could have been renamed freely all along. `RCUserId` stays: it drives the Matrix room lookup. `RCToken` is dead and made optional in ORISO-UserService#751.

Storypapst:
@coderabbitai review

Please re-review — this PR changed materially since the last look. It now also fixes six components that the route-param rename had silently broken (`useParams<{ rcGroupId }>()` readers left behind by the `:groupId` rename), plus a test that mocked the old param name and therefore kept passing against it. 264 tests pass across the affected component trees.

Issue #232, opened 2026-06-23
FE-M10:  DevToolbar in production bundle: 2FA requirement and E2EE toggleable per localStorage flag


Shirloin:
### Affected Service/Component

Frontend (ORISO-Frontend)

### Priority

P2-Medium

### Bug Description

Two `DevToolbar` localStorage flags have no `NODE_ENV` guard at their consumption sites and affect production behavior even though the toolbar UI is hidden by default in production:

1. `disable 2fa_duty` (actual key: '`disable 2fa_duty`' with a space) — read in `TwoFactorNag.tsx` and `TwoFactorAuth.tsx` without any environment check. Setting `localStorage.setItem('disable 2fa_duty', '1')` in a production browser console allows a consultant to bypass the client-side 2FA setup obligation overlay. Note: Keycloak-enforced login OTP is unaffected — this only bypasses the in-app nag/duty UI enforcement.
2. `attachement_encryption` (note: typo in key name) — read in `MessageAttachment.tsx` and `messageSubmitInterfaceComponent.tsx` without any environment check. Setting `localStorage.setItem('attachement_encryption', '0')` in a production browser console disables client-side attachment encryption on the legacy RocketChat upload path. Matrix-backed session uploads are unaffected. Default is already `'1'` (encryption enabled) via `appConfig.attachmentEncryption`.

Note on `e2ee_disabled`: Despite being listed as a concern, this flag has an explicit `NODE_ENV` guard in `e2eeSettings.ts` (`resolveE2eeDisabledFromLocalStorage` returns `false` in non-development environments) — it is not exploitable in production and does not need to be fixed.

### Steps to Reproduce

1. Open production app in browser
2. Open DevTools → Console
3. Run: `localStorage.setItem('disable 2fa_duty', '1')` → reload → consultant 2FA obligation overlay is bypassed
4. Run: `localStorage.setItem('attachement_encryption', '0')` → upload an attachment in a RocketChat-backed session → attachment uploaded without client-side encryption

### Expected Behavior

Flags that affect security-relevant behavior should either be removed entirely outside development, or have explicit `NODE_ENV` guards at all consumption sites — not just at the DevToolbar UI layer.

#### Solution / What to do instead

1. Add `NODE_ENV` guard at consumption sites for `disable 2fa_duty`:
- `TwoFactorNag.tsx` — wrap the `getDevToolbarOption(STORAGE_KEY_DISABLE_2FA_DUTY)` call
[...]

PR #796, opened 2026-07-28
fix(security): ignore dev-only DevToolbar switches outside development (FE-M10)

kiodreambau:
## Why

FE-M10 (#232) reported that 2FA and E2EE are toggleable per localStorage flag in the production bundle. While verifying it for the security-audit closure pass (OpenResilienceInitiative/ORISO-Docs#72), it turned out the existing mitigation was thinner than it looked:

`DEV_ONLY_SWITCH_KEYS` only fed `getVisibleLocalStorageSwitches()`, which filters what the toolbar *renders*. `getDevToolbarOption` read `localStorage.getItem(key)` unconditionally. So in a production build:

- setting `2fa` to `"0"` suppressed the 2FA obligation dialog — `TwoFactorNag.tsx:40` gates the entire nag on that switch
- setting `e2ee_disabled` still took effect, even though that switch was already considered "gated"

Hiding the control was never the same as disabling it.

## What changed

- New `devToolbarKeys.ts` holds the affected keys plus `DEV_ONLY_SWITCH_KEYS` and the `canUseStoredValue()` guard; `DevToolbar.tsx` re-exports the keys so existing imports keep working
- `getDevToolbarOption` now falls back to the compiled-in default for dev-only switches whenever `NODE_ENV !== 'development'` — the stored value is ignored, not just hidden
- Both 2FA switches (`2fa`, `disable 2fa_duty`) join the E2EE switch in that set

## Verification

```
npx vitest run src/components/devToolbar/devToolbarKeys.test.ts        4 passed  (red before devToolbarKeys.ts existed)
npx vitest run src/components/twoFactorAuth src/components/devToolbar src/globalState   26 passed
npx tsc --noEmit                                                       no errors in the touched files
```

Not verified in the browser preview on purpose: the change is a no-op in a development build, which is the only mode the dev server runs — the behaviour under test is the production path.

Refs OpenResilienceInitiative/ORISO-Frontend#232

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security for development-only toolbar switches by preventing stored security settings from affecting production environments.
  * Preserved access to non-security preferences, such as locale settings, across environments.

* **Tests**
  * Added coverage verifying environment-based handling of development-only switches and security-related settings.

### DIFF
--- a/src/components/devToolbar/DevToolbar.tsx
+++ b/src/components/devToolbar/DevToolbar.tsx
@@ -247,12 +247,8 @@ const LOCAL_STORAGE_SWITCHES: (TLocalStorageSwitches | null)[] = [
 ];
 
 const getVisibleLocalStorageSwitches = () =>
-	LOCAL_STORAGE_SWITCHES.filter(Boolean).filter(
-		(localStorageSwitch) =>
-			!(
-				DEV_ONLY_SWITCH_KEYS.has(localStorageSwitch.key) &&
-				process.env.NODE_ENV !== 'development'
-			)
+	LOCAL_STORAGE_SWITCHES.filter(Boolean).filter((localStorageSwitch) =>
+		canUseStoredValue(localStorageSwitch.key)
 	);
 
 export const useDevToolbar = () => {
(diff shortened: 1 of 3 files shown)


kiodreambau:
Addressed the CodeRabbit warning: `getVisibleLocalStorageSwitches` now calls `canUseStoredValue` instead of repeating the `DEV_ONLY_SWITCH_KEYS` + `NODE_ENV` predicate, so rendering and storage reads share one utility. `vitest run src/components/devToolbar src/components/twoFactorAuth` → 23 passed; eslint and tsc clean on the touched file.

shazia-k (approved):

Issue #859, opened 2026-08-02
E-mail kit: emit per-engine template dialects so Java and Keycloak can consume it


Storypapst:
The kit is TypeScript. None of the services that actually send mail can import it:

| Consumer | Templating |
| --- | --- |
| MailService | Thymeleaf — `th:text`, `th:href` |
| UserService direct-SMTP senders | plain string replacement is fine |
| Keycloak theme | FreeMarker — `${...}` |

So `npm run emails:build` has to emit more than one dialect from the same content model, and the output has to be consumable by a Java build rather than copied by hand.

**Scope**
- Add a dialect option to the renderer: `plain` (current `{{mustache}}`), `thymeleaf`, `freemarker`.
- Emit into `src/emails/dist/<dialect>/<tone>/<id>.{html,txt}`.
- Decide and document how the artefact reaches the other repos — published npm package, committed files consumed by a Helm/CI step, or a git submodule. Whatever we pick has to survive a template change without a manual copy.
- Placeholder names stay stable across dialects, so the catalogue in `Email/Foundations` remains the single source of truth.

This blocks every "→ kit" issue below.

Kit lives in ORISO-Frontend `src/emails/` (branch `feat/email-design-system-storybook`).

Part of #828.

## Reviewer test plan

- [ ] `npm run emails:build` emits all three dialects
- [ ] A Thymeleaf-rendered sample matches the Storybook story byte-for-byte apart from the substituted values
- [ ] A FreeMarker-rendered sample does the same
- [ ] The placeholder catalogue in `Email/Foundations` still lists every placeholder correctly
- [ ] Documented how a downstream repo picks up a new build
- [ ] Checked at 375px and at 320px (narrow phone)

Storypapst:
**Built** — branch `feat/email-design-system-storybook`, commit `029bbbec`, [PR #881](https://github.com/OpenResilienceInitiative/ORISO-Frontend/pull/881).

`npm run emails:build` now emits three dialects from one renderer into `src/emails/dist/<dialect>/<tone>/<id>`:

| Dialect | Consumer | Placeholder |
| --- | --- | --- |
| `plain` | UserService direct senders | `{{name}}` |
| `thymeleaf` | MailService | `[[\${name}]]`, plus a `th:*` twin per attribute |
| `freemarker` | Keycloak theme | `\${(name!'')?html}` |

[...]

Issue #828, opened 2026-07-31
EPIC: transactional e-mails — one design system across every sending path


Storypapst:
Every mail ORISO sends should look like it comes from the same sender, carry the same privacy promises, and be reviewable before it goes out. Today it does not: four independent rendering paths produce four different skeletons, and nobody can see any of them without triggering a real send.

The transactional e-mail design system now exists and is anchored in Storybook under `Email/` — one atomic kit (atoms → molecules → organisms → pages), seven occasions, three tone variants, both MIME parts, 139 stories. This epic is about closing the gap between that kit and what production actually sends.

---

## 1. What is sent today

Inventory taken from `origin/pre-dev` across UserService, Helm, Keycloak and the mail service. Four paths, each with its own skeleton.

### Path A — MailService (Thymeleaf), driven by `EmailNotificationFacade`

Templates live in the mail service under `templates/<name>[.en].html`, German and English.

| Template | Trigger | Recipient |
| --- | --- | --- |
| `enquiry-notification-consultant` | `CreateEnquiryMessageFacade` | counsellors of the agency |
| `direct-enquiry-notification-consultant` | `CreateEnquiryMessageFacade` | the addressed counsellor |
| `message-notification-consultant` | `UserController` / `UserSupportControllerDelegate` | counsellor |
| `message-notification-asker` | same | advice seeker |
| `assign-enquiry-notification` | `AssignSessionFacade` | counsellor |
| `daily-enquiry-notification` | scheduled digest | counsellors |
| `reassign-request-notification` | `UserSupportControllerDelegate` | counsellor |
| `reassign-confirmation-notification` | same | counsellor |
| `feedback-message-notification` | feedback/supervision message | counsellor |
| `free-text` | generic | any |

**`OpenResilienceInitiative` has no mail-service repository.** UserService talks to `http://mailservice.<namespace>:8080/mails/send` (`userservice-configmap-env.yaml`), i.e. the upstream Online-Beratung service running unmodified. Adopting the design system on this path therefore needs a deliberate decision — fork the service, or mount an overriding `templates/` directory into the existing image. That decision belongs in an ADR before any template work starts.

### Path B — direct SMTP from UserService, HTML hand-written in Java

These bypass the mail service entirely and build their own markup inline.

| Sender | Mail |
| --- | --- |
| `MagicLinkLoginService` | magic-link login |
[...]

PR #881, opened 2026-08-02
feat(emails): transactional e-mail design system, anchored in Storybook

Storypapst:
Adds the transactional e-mail design system and anchors every piece of it in Storybook under a dedicated `Email/` tree.

Frontend-only. Nothing in this PR changes what production sends — it establishes the kit that the sending services will adopt, tracked in #828 and its 24 sub-issues.

## What is in here

```
src/emails/kit/       atoms → molecules → organisms → document, plus both renderers
src/emails/content/   copy per tone variant (de-sie, de-du, en)
src/emails/preview/   Storybook-only harnesses (iframe preview, inbox line, token sheet)
src/emails/stories/   Email/Foundations, Atoms, Molecules, Organisms, Pages
src/emails/scripts/   buildEmails.mts
src/emails/dist/      generated, committed — 42 send-ready files
```

Seven occasions × three tone variants × two MIME parts. **139 stories over 25 components.**

Outside `src/emails/` this touches exactly two things: one glob in `.storybook/main.ts`, and one script line in `package.json`.

## Why a separate kit rather than reusing the app component library

E-mail clients cannot read CSS custom properties, so no token can be a variable at runtime — every colour and size has to be inlined as a literal, layout has to be tables, and there is no JavaScript. Same design language, incompatible rendering rules.

Everything is therefore prefixed `email…` and lives under `Email/` in Storybook, so the two libraries cannot be mistaken for each other.

## One content model, two MIME parts

`renderEmailHtml` and `renderEmailText` are driven by the same `EmailContent` object. The plain-text twin can no longer drift from the HTML the way hand-kept pairs always do — and the text part is what a text-only client, a screen reader in text mode, and a suspicious spam filter actually read.

## `dist/` is committed on purpose

`npm run emails:build` writes 42 files with their `{{placeholders}}` intact. Committing the output means that after changing an atom, `git diff` shows every mail that atom touches. That diff is the review surface.

## Mobile was measured, not assumed

All 139 stories are rendered headless and checked for script errors, empty renders and horizontal overflow. Two failed on the first run: `Email/Pages/BeraterinKontakt` overflowed to **391px inside a 320px viewport**, because a counsellor e-mail address is an unbreakable token. Fixed with `word-break` on panel values and the footer contact line. Final state **139/139 clean**.

Beyond that, the responsive stylesheet now:

- releases the fixed 600px column and trims the canvas padding, so a 320px screen spends its width on content rather than margins;
- **actually** stretches the CTA — `display:block` on the anchor alone did nothing while its wrapper table was still shrink-to-fit;
- stacks the data panel's label/value pairs, which otherwise squeezed a postal address into a 38%-wide column;
- lets the four footer links wrap instead of forcing a horizontal scroll;
- resets `x-apple-data-detectors`, so iOS stops recolouring the date and address in the appointment mail into system blue.

The `<style>` block is the only non-inline CSS and exists solely for the `max-width:620px` media query. Clients that strip it — the Gmail Android app on a non-Gmail account, most notably — fall back to the desktop table layout, which still renders. The media query only ever *improves* narrow viewports; it is never load-bearing.

## Copy rules encoded in the kit

- A mail to an advice seeker names no message content, no person and no counselling centre. The two exceptions, `termin` and `beraterin-kontakt`, are allowed only because the recipient triggered them.
- One action per mail.
- The subject and preheader are the risky part — they show up on a lock screen. Neither reveals the topic. `Email/Atoms/Preheader → Every mail` shows all seven side by side for exactly this check.
- The encryption promise closes every card, because the mail is the one part of the product that is not encrypted and not behind a login.

## Known limitation

The kit is TypeScript, and none of the Java services that actually send mail can import it. `emails:build` currently emits one dialect (plain `{{mustache}}`); MailService needs Thymeleaf and the Keycloak theme needs FreeMarker. That is #859 and it is the enabler for most of the epic — worth knowing while reviewing this, but out of scope here.

## Reviewer test plan

Run `npm run storybook` and open the `Email/` tree.

- [ ] `Email/Foundations → Catalogue` lists all seven mails with their audience and placeholders
- [ ] `Email/Pages/Termin` renders in all three tone variants and the copy reads naturally in each
- [ ] `Email/Pages/Termin → Narrow phone (320px)` — data panel stacks, button spans the column, no horizontal scrolling
- [ ] `Email/Pages/BeraterinKontakt → Narrow phone (320px)` — the counsellor e-mail address wraps instead of overflowing
- [ ] `Email/Pages/…  → text/plain part` says the same thing as the HTML for at least two mails
- [ ] `Email/Atoms/Preheader → Every mail` — no subject or preview line reveals a topic, a name or a counselling centre
- [ ] `Email/Organisms/Shell → Other tenant brand` — colours and platform name change throughout, nothing stays hardcoded
- [ ] `Email/Atoms/LogoLockup → Images blocked` — the platform name is still readable as text
- [ ] Any page story → `Template file (placeholders)` shows `{{…}}` rather than blanks
- [ ] `npm run emails:build` leaves `git status` clean (the committed `dist/` matches the kit)
- [ ] Checked at 375px as well as 320px

Refs #828


Storypapst:
## Verification evidence

|---|---|---|---|
| INFORMATIONAL | Local | `0f22ac8` | Claude |

### The 'neue Nachricht' mail as it arrives — Email/Pages/NeueNachricht, Deutsch \(Sie\)\. No content, no name, no counselling service\.

The 'neue Nachricht' mail as it arrives — Email/Pages/NeueNachricht, Deutsch \(Sie\)\. No content, no name, no counselling service\.

### The 'neue Nachricht' mail as it arrives — Email/Pages/NeueNachricht, Deutsch \(Sie\)\. No content, no name, no counselling service\.

The 'neue Nachricht' mail as it arrives — Email/Pages/NeueNachricht, Deutsch \(Sie\)\. No content, no name, no counselling service\.

### The 'neue Nachricht' mail as it arrives — Email/Pages/NeueNachricht, Deutsch \(Sie\)\. No content, no name, no counselling service\.

The 'neue Nachricht' mail as it arrives — Email/Pages/NeueNachricht, Deutsch \(Sie\)\. No content, no name, no counselling service\.

[All evidence for this run](https://evidence.dreambau.com/r/w4dtd3a526sex37raqtd7xa32qhojc7e)

Storypapst:
## Verification evidence

|---|---|---|---|
| INFORMATIONAL | Local | `0f22ac8` | Claude |

### The 'neue Nachricht' mail as it arrives\. No message content, no name, no counselling service — that is the anonymity rule of ADR\-019, rendered\.

The 'neue Nachricht' mail as it arrives\. No message content, no name, no counselling service — that is the anonymity rule of ADR\-019, rendered\.

[All evidence for this run](https://evidence.dreambau.com/r/yvwcvxxmoc7y27aqin4gv54wv76wdgvq)

Storypapst:
## Verification evidence

|---|---|---|---|
| INFORMATIONAL | Local | `0f22ac8` | Claude |

### Advice seeker: three switches, plus a named list of what is sent regardless\.

Advice seeker: three switches, plus a named list of what is sent regardless\.

[All evidence for this run](https://evidence.dreambau.com/r/bmsydmkrmvekssdsdg4ukhjddismsuck)

Storypapst:
## Verification evidence

|---|---|---|---|
| INFORMATIONAL | Local | `0f22ac8` | Claude |

### Counsellor: seven switches over the same catalogue\. Same occasions, a different list — the point of ADR\-019\.

Counsellor: seven switches over the same catalogue\. Same occasions, a different list — the point of ADR\-019\.

[All evidence for this run](https://evidence.dreambau.com/r/kbzi2sda2xsia4z5xz8ohu8t4csdkr4u)

Storypapst:
## Verification evidence

|---|---|---|---|
| PASS | Local | `a6aa2b3` | Claude |

### Email/Pages/Termin \(de\-du\), Sample data ON — real values render

Email/Pages/Termin \(de\-du\), Sample data ON — real values render

### Email/Pages/Termin \(de\-du\), Sample data ON — real values render

Email/Pages/Termin \(de\-du\), Sample data ON — real values render

### Email/Pages/Termin \(de\-du\), Sample data ON — real values render

Email/Pages/Termin \(de\-du\), Sample data ON — real values render

### Email/Pages/Termin \(de\-du\), Sample data ON — real values render

Email/Pages/Termin \(de\-du\), Sample data ON — real values render

### Email/Pages/Termin \(de\-du\), Sample data ON — real values render

Email/Pages/Termin \(de\-du\), Sample data ON — real values render

### Email/Pages/Termin \(de\-du\), Sample data ON — real values render

Email/Pages/Termin \(de\-du\), Sample data ON — real values render

[All evidence for this run](https://evidence.dreambau.com/r/dpaig5hos3q5xzrnwunwx5pntka854ut)

Storypapst:
@Shirloin thanks — both points addressed.

**1. Undeclared production change.** You are right that the description undersold this. Taken the second option you offered: the PR description is rewritten and now opens with an explicit scope section declaring the EmailNotifications settings rewrite (ADR-019 role-split matrix), the three new DTO fields, and the hard dependency on UserService #958 for persisting them — including what happens until that lands (the three affected switches render but do not persist). Splitting the screen out would have left the kit's `?mail=<occasion>` deep links and the unsubscribe footer pointing at a screen that does not exist on this branch, so declaring the dependency and landing them together on pre-dev seemed the more honest cut.

**2. Stale documentation and the CSS typo.** Both fixed at the source in the follow-up commit:
- `mso-line-height-rule:exactly;;` came from `emailBodyTextStyle()` ending in `;` while `emailBlock` appends its own separator. One character removed in `emailAtoms.ts`, dist regenerated — the 214-file dist diff is exactly that one character everywhere, which is the committed-dist review model working as intended.
- README and story prose no longer say "seven occasions"; the catalogue has 22, and running text is now count-agnostic where a hardcoded number would just go stale again.

Gates after the fix: eslint `--max-warnings=0` + tsc, Storybook typecheck, 3184 unit tests, Storybook build, and a headless sweep of all 280 `Email/*` stories (every page story additionally forced to 320px, 456 renders total) — all green. Fresh screenshots are in the updated description.

PR #853, opened 2026-08-01
fix(matrix): recover from stale crypto device conflicts

Storypapst:
## Summary

- observe Rust-Crypto child logger failures without changing SDK log output
- detect the exact stale-device one-time-key conflict from #551
- clear only the persisted Matrix device id, request a fresh device-bound token, and reinitialize crypto once
- preserve anonymous-session crypto isolation semantics across recovery

## PreDev reproduction

Apu on PreDev repeatedly receives `400 M_UNKNOWN` from `/_matrix/client/v3/keys/upload`: the same `signed_curve25519` key id already exists with different key material. Both sender and recipient then see only “Message encrypted”, including newly sent messages.

## Verification

- `npm run test:unit -- src/services/matrix*.test.ts src/components/sessionCookie/getMatrixAccessToken.test.ts src/utils/matrixLogging.test.ts` — 97 passed
- `npx tsc --noEmit`
- targeted ESLint on all six changed files
- `git diff --check`

Browser acceptance remains pending regular review, merge, and PreDev deployment.

Fixes #551

Issue #551, opened 2026-07-19
Recover from one-time-key upload conflict — stale device_id with lost crypto store wedges all encrypted sends


kiodreambau:
## Problem

When the stored Matrix `device_id` (localStorage `matrix_device_id:{userId}`) survives but the crypto store (IndexedDB) does not — Safari storage eviction, cleared site data, restored sessions, second-browser relogin patterns — the rust-crypto stack regenerates one-time keys for the SAME device and the upload fails forever:

```
Failed to process outgoing request 0: M_UNKNOWN: MatrixError: [400]
One time key signed_curve25519:AAAAAAAAAA0 already exists.
```

The outgoing crypto request queue wedges on this request; encrypted sends and key-shares silently stop working. The composer shows a local echo, the user believes the message was sent. Reproduced repeatedly on pre-dev during live-chat testing 2026-07-19 (consultant device restored from cookies+localStorage only).

## Expected behavior

- Detect the OTK conflict (`M_UNKNOWN` + "already exists" on `/keys/upload`) and recover automatically: drop the stored device id, re-login with a fresh device (or wipe and re-initialize the crypto store for the stored device).
- Surface a user-visible failed-delivery state instead of a silent local echo (relates to #542 / #537 delivery status).

## Related

#436 (crypto track), #438 (invisible crypto), #439 (dehydrated devices), #542 (encryption-broke delivery status).

Storypapst:
Additional PreDev impact proven in the same Homer → Apu session on 2026-08-01:

- Apu started a real 1:1 video call with fake camera/microphone permissions.
- Homer received no actionable incoming-call UI.
- The room gained only another encrypted placeholder event.
- Instrumented result: `incomingVisible=false`, zero `RTCPeerConnection` instances on both participants.
- `/keys/upload` continued returning the exact OTK conflict documented above.

This means the stale-device crypto wedge blocks not only readable chat/key sharing but also encrypted 1:1 call signalling. PR #853 should therefore be browser-accepted with both a readable message and a paired audio/video call after regular PreDev deploy.

Storypapst:
Post-merge PreDev acceptance on regular Frontend digest `sha256:111de1ce...` found a remaining recovery race:

- #853 successfully detects the exact OTK conflict, rotates both restored browser states to fresh device IDs, and subsequent `/keys/upload` completes with HTTP 200.
[...]

PR #885, opened 2026-08-02
fix(call): enable per-participant media E2EE by default

nikunjdecyb:
## Summary
- Restore ADR-018 default: Element Call widgets request `perParticipantE2EE=true` so LiveKit media is end-to-end encrypted
- Keep `releaseToggles.enableCallMediaE2EE === false` as an emergency kill-switch (ORISO-ElementCall#35 silent no-audio)
- Update widget unit tests for default-on / kill-switch-off paths

## Context
After PR #832, media E2EE was opt-in and unset in settings, so calls correctly showed **Not encrypted** because media was genuinely unencrypted. Signalling/room Megolm was unrelated. Issue #551 (stale device/OTK) is already on `pre-dev` via #853/#883 and is out of scope here.

## Test plan
- [x] `vitest` `src/components/call/widget/useElementCallWidget.test.tsx` (11 passed)
- [ ] PreDev two-browser video call: encrypted indicator + audio/video both directions
- [ ] Kill-switch: `enableCallMediaE2EE: false` omits `perParticipantE2EE` if needed
- [ ] Confirm deployed Element Call honours host `perParticipantE2EE` (EC#36)

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Media end-to-end encryption is now enabled by default for Element Call.
  * Transport-only media is used only when the encryption setting is explicitly disabled.

* **Documentation**
  * Clarified the encryption setting’s default behavior and fallback when key distribution fails. Call signalling and room events remain encrypted.

### DIFF
--- a/src/components/call/widget/useElementCallWidget.test.tsx
+++ b/src/components/call/widget/useElementCallWidget.test.tsx
@@ -164,17 +164,15 @@ describe('useElementCallWidget', () => {
 		expect(fragment.has('accessToken')).toBe(false);
 		expect(fragment.has('password')).toBe(false);
 		expect(fragment.has('enableE2EE')).toBe(false);
-		// Media E2EE is opt-in per environment: the media keys ride the MatrixRTC
-		// to-device transport, so a gap in the host's Olm path would connect the
-		// call with no audio and no error (ORISO-ElementCall#35).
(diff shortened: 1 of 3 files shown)


shazia-k (approved):
LGTM — clean change, tests and CI look good. Approving and merging.

PR #886, opened 2026-08-02
fix(call): release Element Call warm-up MediaStreams

nikunjdecyb:
## Summary
- 1:1 session call UI warmed up `getUserMedia` and stored the stream on `window.__preRequestedMediaStream` for the legacy native widget.
- Outgoing calls always use Element Call (`usesElementCall: true`), so FloatingCallWidget never released that stream; `endCall` only cleared `__activeMediaStream`.
- Orphaned `live` tracks left the camera busy and broke re-activation after leave/refresh.
- Fix: stop warm-up tracks immediately after permission grant; release both window streams on hangup/reject; shared cleanup helper used by SessionMenu, CallManager, and call widgets.

## Test plan
- [x] `npx vitest run src/utils/callMediaStreamCleanup.test.ts` — PASS
- [x] Local Playwright old-vs-new leak repro — old path leaves `readyState: live`; new path `ended`
- [x] ESLint on touched files — PASS
- [ ] Manual on predev: join video call → leave → refresh → join again; camera activates each time; no stuck camera LED
- [ ] Confirm 1:1 and group call hangup still tears down Element Call iframe

## Risks / follow-ups
- Branch may need rebase onto latest `origin/pre-dev` before merge (created from an older local tip).
- Full headless predev call loop was blocked by 2FA setup redirect on the test account; please confirm manually after dismissing 2FA.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved call media cleanup to prevent orphaned camera and microphone tracks.
  * Reduced the risk of failed re-joins and lingering media access when starting, ending, rejecting, or switching calls.
  * Improved cleanup behavior across one-to-one and group calls, including mobile call flows.

* **Tests**
  * Added coverage for stopping tracks and releasing temporary and active call media streams.

* **Documentation**
  * Documented media-stream cleanup guidance and related call-flow findings.


shazia-k (approved):
LGTM — fixes the orphaned warm-up MediaStream leak on the Element Call path. Shared cleanup + tests look good, CI green

Issue #892, opened 2026-08-02
Chat message components: four design deviations found via Storybook


Storypapst:
> **Status · 12.08.2026 — all four delivered.** Deviations 1–3 are on `pre-dev` via PR #905 and #945. The icon-only reject button and the Tigrinya consent string are in PR #1043, awaiting review. That PR also translates the three consent strings, which existed only in German.

---

**Affected repos:** ORISO-Frontend

## Why

Rendering every chat message component in Storybook (OpenResilienceInitiative/ORISO-Frontend#826, PR OpenResilienceInitiative/ORISO-Frontend#849) surfaced four deviations in **shipped** components. They are grouped here because they are small, related and all live in the message/dialog layer — keeping them out of #849 so that the Storybook coverage can merge on its own.

All four verified against `origin/pre-dev`.

### 1. The visibility chip must not appear when everyone can see the message

`MessageItemComponent.tsx:1386`:

```ts
const showVisibleAudience =
    visibleAudienceSummaryLabels.length > 0 &&
    !isDeleteMessage && !isSystemNotification && !alias?.messageType;
```

No guard for the all-audience case. `visibilityGroups` already computes `isAllAudienceSelected` at `:701` — including `assumeAllAudienceByDefault` when `visibleToUserIds` is empty in a group — but that value never reaches the render condition. Today a group message renders a chip reading "visible only to: **Alle**", which states the opposite of what a restriction chip is for.

The chip exists to mark a message that **only some** participants can see. Hide it whenever the audience is everyone.

### 2. Carimat's avatar must use the message-avatar geometry

Carimat is a message type like any other system message — a child of the chat message, not a sibling. Its circle must therefore take the same parameters as `MessageAvatar` (`size = 32`), instead of its own 54×54 frame in a 64px column (`PseudonymCard.styles.scss:34-40`).

### 3. The system notice uses the wrong pattern

`MessageSendFailed` is the correct one: icon in a circle, title, subtitle, bubble, timestamp.

The system-notification path renders a "Systembenachrichtigung" chip **plus** a "Systemhinweis" headline **plus** a robot avatar — three labels competing to say the same thing. Align it with `MessageSendFailed`. Reference: [App.Oriso 8607-28488](https://www.figma.com/design/L2mOFNSGdxPPx1XA4HFAog/App.Oriso?node-id=8607-28488&m=dev).

### 4. The consent dialog deviates from its design, and breaks at 375px

[...]

PR #905, opened 2026-08-03
Fix three chat design deviations: visibility chip, Carimat avatar, consent dialog

Storypapst:
Closes OpenResilienceInitiative/ORISO-Frontend#892 — partially; deviation 4 stays open, see below.

**Based on `feat/storybook-message-variants` (PR OpenResilienceInitiative/ORISO-Frontend#849), not on `pre-dev`** — the story updates here build on the ones added there. Merge #849 first and this diff collapses to just these fixes.

## 1. The visibility chip claimed a restriction that did not exist

A normal group message rendered **"visible only to: Alle"** — a restriction chip stating there is no restriction.

`isAllAudienceSelected` was already being computed inside `visibilityGroups` (including the implicit case: empty recipient list in a group = everyone), but the value never reached the render condition. Lifted into its own memo and added to the guard.

Two stories pin both directions, so the fix cannot swing the other way and hide the chip everywhere:

- `Group message everyone can see — NO visibility chip`
- `Group message restricted — chip IS shown`

## 2. Carimat's avatar had its own geometry

Carimat is a *kind of chat message*, so its circle has to match `.messageItem__avatar` — **60×60 with a 6px white ring**. It carried a bespoke **54×54** frame in a 64px column, which set Carimat rows a few pixels off from the human rows beside them in the same conversation.

## 3. The consent dialog named the wrong modality and broke on small phones

| | Before | After |
|---|---|---|
| Modality | "einen **Video-Call** mit Ihnen starten" | "einen **Chat** mit Ihnen beginnen" |
| Wording | "Beraterinnen oder Berater" | "eine beratende Person" (ADR-018 §7) |
| Headline | "Herzlich Willkommen**!**" | "Herzlich Willkommen" |
| Buttons ≤575px | one row, broke out of the card | stacked, full width, inside the card |
| Headline ≤575px | "Herzlich Willkomm\|en!" | wraps between words |

**"Video-Call" was in all seven locales** — the error had been translated through. Corrected in `de`, `de@informal`, `en`, `fr`, `ru`, `tr`.

**Tigrinya is deliberately untouched** and still says video call. Editing safety-relevant consent copy in a language nobody on the team reads would be guessing; it needs a native speaker. Flagged rather than silently machine-edited.

Root causes of the layout break, for the record: the existing `640px` block only shrank button padding, which was never going to be enough for two labelled buttons; `sanitize.css` sets a global `word-break: break-word`, which is what split the headline mid-word; and 24px outer plus 32px inner padding consumed 112px of a 375px screen.

## Verification

- `tsc --noEmit --project tsconfig.storybook.json` — clean
- `npm run build-storybook` — exit 0
- `node scripts/shoot-stories.mjs` — **125 stories, 0 problems**
- The 375px story now renders at a **real 375px viewport**, not a shrunk container, so the component's own media queries actually fire. The script was changed to give width-named stories that width as a viewport — worth noting, because the earlier version of this claim was measured wrongly.

## Not included

**Deviation 4 — the system notice** renders a "Systembenachrichtigung" chip *and* a "Systemhinweis" headline *and* a robot avatar, where `MessageSendFailed` uses one label. Deciding which of the three survives is a design decision, not a mechanical fix, so it stays open on OpenResilienceInitiative/ORISO-Frontend#892.

## Reviewer test plan

- [ ] `Components/Chat/MessageItem → Group message everyone can see` shows **no** chip
- [ ] `Components/Chat/MessageItem → Group message restricted` still shows the chip
- [ ] `Components/Chat/CarimatMessage → Sequence` — Carimat's circle lines up with the message avatars in `MessageItem`
- [ ] `Components/Dialog/AnonymousConsentGate → Mobile 375px` — buttons stacked inside the card, headline unbroken
- [ ] The dialog says "Chat", not "Video-Call", in de / de@informal / en / fr / ru / tr
- [ ] **Mobile:** check the dialog on a real phone at 375px and 390px
- [ ] Tigrinya still says video call — expected, needs a native speaker


Storypapst:
## Deviation 4 is now included — all four are fixed

Took the Figma reference ([App.Oriso 8607-28488](https://www.figma.com/design/L2mOFNSGdxPPx1XA4HFAog/App.Oriso?node-id=8607-28488&m=dev)) rather than deciding the presentation myself. It shows the canonical system message as **circle icon + bold title + quiet subtitle + body bubble** — which is exactly what `MessageSendFailed` already renders. So this aligns the generic system notification with the existing pattern instead of inventing a third one.

**Before:** one notice wore three labels at once — a "Systembenachrichtigung" chip, a 24px "Systemhinweis" headline, and a description, all inside the bubble, with the robot avatar beside them.

**After:** title in the header (bold), the "System Notification" qualifier as the quiet subtitle instead of a chip, description alone in the bubble. Reusing the `sendFailed` classes on purpose — one system-notice presentation that cannot drift into two.

### One thing worth recording

`MessageSendFailed` puts its avatar in a real side column, so its header needs no indent. In `MessageItemComponent` the avatar **deliberately overlaps** the content column — see the 19px-overlap note in `message.styles.scss` — which is why the name line carries `padding-left: 20px`.

Dropping the header in without that clearance ran the title under the avatar. Measured before the fix: avatar at x=16–48, header starting at x=33, so **15px of overlap** with the title visibly clipped. Added `__systemNotificationHeaderText` for the same clearance.

Also removed the now-orphaned `__systemNotificationTag` and `__systemNotificationTitle` rules rather than leaving dead CSS behind.

### Verification

- `tsc --noEmit` clean · `build-storybook` exit 0 · **125 stories rendered, 0 problems**
- `MessageSendFailed` renders unchanged — checked explicitly, since this reuses its classes

The PR body still says deviation 4 is out of scope; that is now stale — it is in.

Storypapst:
## Verification evidence

|---|---|---|---|
| PASS | Local | `15c282a` | Claude |

### BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

### BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

### BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

### BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

### BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

### BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

### BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

### BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

BEFORE — the bug: a normal group message renders the restriction chip 'visible only to: Alle', i\.e\. a restriction chip stating there is no restriction\. Plus a '\+0' recipient badge\.

[All evidence for this run](https://evidence.dreambau.com/r/w3fht62y28tg5wmfvadrhrh64gthagw7)

shazia-k (approved):

Issue #964, opened 2026-08-03
[FE-H05a] Remove Caritas legal-URL fallback, branding and storage keys from the frontend


kiodreambau:
Part of #178

## Why

`runtimeConfig.ts` is already environment-driven, but the compliance defect this
epic reports still has a live path: `TermsAndConditions.tsx` reads
`process.env.REACT_APP_LEGAL_PRIVACY_URL` directly and falls back to a
third-party Caritas privacy page. Under KDG/GDPR that is a wrong mandatory
provider disclosure, not cosmetic.

Alongside it, the app still ships Caritas branding: the login stage renders the
seven-logo Caritas-consortium block, four localStorage keys and DOM events are
namespaced `caritas_*` / `caritas:*`, and `.env.example` documents Caritas
domains as the example values.

`scripts/validate-hardcoded-hosts.sh` already greps the build output for
`caritas.de` but is wired into nothing, so nothing stops these from coming back.

## What to build

1. Route `TermsAndConditions.tsx` through `getLegalPrivacyUrl()` from
   `resources/scripts/runtimeConfig` — delete the hardcoded Caritas URL and the
   direct `process.env` read.
2. Rename the live-chat storage keys and events from `caritas_*` / `caritas:*`
   to `oriso_*` / `oriso:*`, migrating any existing value so consultants do not
   silently lose their availability preference.
3. Replace the hardcoded seven-logo block in `stage.tsx` with the tenant's
   configured theming logo; render no logo row when the tenant has none. Delete
   the seven charity SVG assets.
4. Replace the Caritas example values in `.env.example` with neutral
   placeholders.
5. Wire `scripts/validate-hardcoded-hosts.sh` into the build so a Caritas domain
   can never re-enter the bundle, and add a source-level regression test.

## Acceptance

- [ ] No Caritas domain appears in any resolved legal URL at runtime.
- [ ] `git grep -i caritas -- src/` returns no branding, storage-key or URL hit.
- [ ] An existing `caritas_liveChatAvailability` value still resolves after the rename.
- [ ] The login stage renders the tenant logo, or no logo row when unset.
- [ ] The hardcoded-host guard runs as part of the build and fails on a seeded Caritas URL.

kiodreambau:
Delivered via #966, merged into `pre-dev` at 2026-08-04 12:13 (merge commit 09adecad).

[...]

Issue #178, opened 2026-06-17
[FE-H05] Caritas legacy remnants: wrong privacy policy / imprint as fallback, Caritas logo, caritas namespace


shanzaeimran:
## 🐛 Finding FE-H05 — Compliance / High (KDG/GDPR)

> **Severity:** High | **Effort:** S | **Repo:** ORISO-Frontend | **Category:** Hardcoded URL / Compliance

## Problem

When build-time environment variables are missing, the application silently falls back to
**third-party Caritas legal pages** as the mandatory privacy policy and imprint.
Under KDG/GDPR this is a wrong mandatory provider disclosure — not cosmetic.

Beyond the legal fallbacks, Caritas branding is scattered across code and infrastructure:

| Location | Line(s) | Remnant |
|---|---|---|
| `src/resources/scripts/config.ts` | 22–29 | Fallback URLs: `caritas.de`, `caritas-beratungundhilfe.de/impressum`, `.../datenschutz` |
| `src/components/termsandconditions/TermsAndConditions.tsx` | 74–76 | Hardcoded duplicate of the privacy fallback URL inside `transformText2DataPrivacyLink()` |
| `src/components/stage/stage.tsx` | 19 | `import { ReactComponent as CaritasLogo } from '...02_caritas.svg'` |
| `src/utils/liveChatToggle.ts` | 13–14 | Storage keys `caritas_liveChatAvailability` / `caritas:liveChatAvailabilityChange` |
| `deployment-v2.yaml` | 5 | `namespace: caritas` |
| `k8s-temp-frontend.yaml` | 5, 39, 51 | `namespace: caritas` in deployment, service, ingress |
| `ingress.yaml` | 5 | `namespace: caritas` |
| `storybook-deployment.yaml` | 5 | `namespace: caritas` |

## Solution

Remove all Caritas fallback values and throw a build error when legal URL env vars are
absent. Migrate the logo to a neutral/tenant-specific asset. Rename storage keys.
Switch all k8s manifests from `namespace: caritas` to `namespace: oriso` (or a Helm value).

## Work Steps

1. **Tests first (red).** Add a build-time validation script `scripts/validate-env.js`
   and a Jest test `src/__tests__/config.legalUrls.test.ts`:
   - Assert `yarn build` exits non-zero when `REACT_APP_LEGAL_PRIVACY_URL` is empty.
   - Assert `config.legalPrivacyUrl` and `config.legalImprintUrl` never contain
[...]

PR #966, opened 2026-08-04
chore(compliance): remove Caritas legacy from shipped frontend code

kiodreambau:
## Why

FE-H05 (#178). When the legal-URL env vars were missing, the app served a
**third party's** privacy policy and imprint as its own mandatory disclosure —
wrong provider information under KDG/GDPR, not a cosmetic issue. Around it,
Caritas branding was still scattered through code a user actually sees.

The `config.ts` fallbacks named in the issue were already fixed on `pre-dev`.
This PR clears what was left.

## What changed

| Leak | Where | Now |
|---|---|---|
| Caritas privacy policy as fallback | `TermsAndConditions.tsx:76` | resolves via `getLegalPrivacyUrl()`; fallback is this platform's own `/datenschutz` |
| Seven hardcoded Caritas-association logos on the login stage | `stage.tsx` | tenant's `theming.associationLogo` (falling back to `theming.logo`); **nothing** when the tenant configured none |
| PWA installs as "Caritas Online-Beratung" | `public/manifest.json` | "ORISO Online-Beratung" |
| Maintenance page button "Zur Caritas Website" | `src/pages/under-construction.html` | "Zur Startseite" |
| Matrix device announced as "Caritas Frontend" | `matrixRegistrationService.ts` | "ORISO Frontend" |
| `caritas_*` / `caritas:*` localStorage keys and events | `liveChatAvailabilityStorage.ts`, `liveChatToggle.ts` | `oriso_*` / `oriso:*`, legacy keys read once then dropped |
| Unreferenced mail template with Caritas logo and "Ihr Caritas Team" | `src/components/mailTemplate/` | removed |
| Now-unused association logo assets | `src/resources/img/logos/` | removed |

A guard test — `src/utils/noCaritasLegacy.test.ts` — scans shipped code for
Caritas legal domains, the logo asset, and Caritas-namespaced storage. The
`LEGACY_*` migration constants are the one documented exception.

## Deliberately NOT done: the `namespace: caritas` rename

The issue asks to switch the manifests to `namespace: oriso`. **That would break
the deploy.** The workloads really do run in the `caritas` namespace — both
`deploy-development.sh` and `deploy-production.sh` talk to `-n caritas`, and the
in-cluster service DNS in the manifests resolves through
`*.caritas.svc.cluster.local`. Renaming the YAML without migrating the cluster
changes nothing except which namespace `kubectl apply` writes into.

Instead, the deploy scripts now read `ORISO_K8S_NAMESPACE`, defaulting to the
[...]

PR #970, opened 2026-08-04
fix(compliance): remove Caritas legal-URL fallback, branding and storage keys

kiodreambau:
## Why

The compliance core of #178 was already fixed — `runtimeConfig.ts` is
environment-driven and falls back to the deployment's **own** origin. But one
live path still bypassed it:

```ts
// TermsAndConditions.tsx:76 (before)
const privacyHref =
    process.env.REACT_APP_LEGAL_PRIVACY_URL ||
    'https://www.caritas-beratungundhilfe.de/datenschutz';
```

That is the mandatory privacy disclosure pointing at a third party — a wrong
mandatory provider disclosure under KDG/GDPR, not a cosmetic default.

## What changed

**The compliance defect**
- `TermsAndConditions` now calls `getLegalPrivacyUrl()`, so every legal link
  resolves through one place instead of reading `process.env` on its own.

**Branding remnants**
- The login stage rendered a fixed seven-logo Caritas-consortium row. It now
  renders the tenant's configured theming logo (via `resolveTenantMediaUrl`, so
  `/media/{id}` uploads resolve on the app origin), and no logo when a tenant
  has none. The seven SVGs are deleted.
  The wrapper element is kept even with no logo — it doubles as the `flex: 1 1 0`
  spacer holding the headline and the legal links apart, so removing it would
  have shifted the stage layout.
- Four live-chat `localStorage` keys / DOM events renamed `caritas_*` -> `oriso_*`.
- `.env.example` gets neutral placeholders plus a note on why these must be the
  operator's own pages.
- Matrix device display name `"Caritas Frontend"` -> `"ORISO Frontend"` (visible
  to users in their Matrix session list).

**Migration** — consultants who set a live-chat preference before the rename
still carry the old key. Both readers adopt it once and retire it, so nobody is
silently dropped out of live-chat routing. Covered by
`liveChatStorageMigration.test.ts`, including the "switched off" case that must
*not* be resurrected.

## Regression guards

Nothing stopped any of this from coming back, so this PR adds two gates.

1. `src/__tests__/caritasRemnants.test.ts` — fails on any third-party Caritas
   URL, `caritas`-namespaced storage key, or consortium logo asset under `src/`.
   Deliberately **not** a blanket ban on the word "Caritas": translation copy
   and demo agency fixtures still legitimately name real organisations, and
   rewriting that copy is a separate content decision.

2. `scripts/validate-hardcoded-hosts.sh` already existed — but was **wired into
   nothing**, and would not have caught this defect anyway: its fixed string
   `caritas.de` is not a substring of `caritas-beratungundhilfe.de`. This PR
   adds a domain regex that does match, and runs the script from `postbuild`, so
   the existing CI build step enforces it.
   Source maps are excluded from the scan — `sourcesContent` embeds original
   comments, so the `app.oriso.org -> api.oriso.org` example in `runtimeConfig.ts`
   tripped the gate on prose that never becomes a runtime value. A genuinely
   hardcoded value is emitted into the `.js` as well, so no real coverage is lost.

## Verification

```
npm run build            # build + postbuild guard: exit 0
npx vitest run           # 232 files, 1533 tests passing
npm run lint:scripts     # eslint --max-warnings=0 && tsc: clean
npm run lint:style       # stylelint: clean
```

The guard was proven to actually fail, not merely pass — seeding either real
Caritas legal URL into a built bundle chunk makes it exit 1, and it returns to
0 once removed:

```
$ printf '\nvar x="https://www.caritas-beratungundhilfe.de/datenschutz";\n' >> build/static/js/109.*.chunk.js
$ bash scripts/validate-hardcoded-hosts.sh build; echo $?
Third-party domain found in build: /caritas[a-z0-9-]*\.(de|org|com|net)/i
1
```

## Review note

The stage logo is the one user-visible change — the login/loading screen goes
from seven fixed charity logos to the tenant's own logo (or none). Flagging it
for a design look, per the issue's "neutral/tenant-specific asset" wording.

## Not in scope

- The **namespace** half of #178 is a separate PR (#969), since it needs
  deployment review rather than frontend review.
- @BjoernLudwig's open question on the issue — whether platform privacy/imprint
  URLs should become admin-panel fields rather than env config — is **not**
  preempted here. This PR only removes the third-party fallback; whichever way
  that decision goes, the URLs still resolve through `runtimeConfig`.

Closes OpenResilienceInitiative/ORISO-Frontend#964
Refs OpenResilienceInitiative/ORISO-Frontend#178


kiodreambau:
Closing as a duplicate of #966, which was opened in parallel for the same issue and covers strictly more surface (PWA manifest, maintenance page, i18n copy, mail-template removal) with an equivalent — and better-scoped — source guard.

Nothing here is lost. The pieces #966 did not have have been ported onto its branch:

- `scripts/validate-hardcoded-hosts.sh` wired into `postbuild`, so CI's existing build step gates the **built output**, complementing #966's source-level `noCaritasLegacy.test.ts`.
- A fix to that script's pattern. It already listed `caritas.de`, but as a **fixed string** — which does not match `caritas-beratungundhilfe.de`, i.e. the exact domain FE-H05 was about. Replaced with a domain regex, and source maps are excluded because `sourcesContent` matches on comments rather than runtime values.
- `.env.example`: neutral placeholders plus a note on why these must be the operator's own pages.
- `legalUrls.test.ts`, pinning the positive contract — legal URLs fall back to the deployment's **own** origin. #966's guard proves the absence of Caritas domains; this proves what replaces them.

The k8s half stays separate in #969, which has been reworked to complement #966 rather than conflict with it: it no longer touches the deploy scripts and now uses #966's `ORISO_K8S_NAMESPACE` name.

PR #987, opened 2026-08-07
fix(session-list): rename the modality label from "Nähe" to "Mail"

### LINKED ISSUE #985: "Nähe" needs to be renamed to "Mail"
> **Status · 11.08.2026** — **Fixed.** PR #1002, merged into `pre-dev` (`c20636ff`). Needs deploy and retest on Pre-Dev.

---

### Affected Service/Component

Frontend (ORISO-Frontend)

### Priority

P0-Critical

### Bug Description

What is called "Nähe" in several menus needs to be renamed to "Mail". "Nähe" is simply to foreign for basically all our practicioners. 

### Steps to Reproduce

1. Go to message list for a counsolor and observe "Nähe" as one if the filter items.

### Expected Behavior

Filter items and all other "Nähe" mentions show "Mail".

### Actual Behavior

see above

### Environment

_No response_

### Relevant Logs/Error Messages

```shell

```

### Screenshots/Videos

_No response_

### Workaround

_No response_

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before
- [x] I have provided all relevant information above
- [x] I can reproduce this bug consistently

Storypapst:
Closes #985

## What changed

Counselling practitioners rejected "Nähe" as too foreign a name for the asynchronous agency modality, so every place that labels that modality now reads **Mail**:

- `sessionList.toolbar.chips.nearby` and `sessionList.toolbar.chips.chats` in **all** locale bundles (de, de@informal, en, fr, ru, ti, tr).
- The hardcoded fallbacks that render when a translation fails to load: the filter chip definition in `SessionsListToolbar.tsx`, the enquiry filter chip in `EnquiryFilterChips.tsx`, and the icon label plus its `alt` text in `SessionListItemComponent.tsx`.
- Storybook mocks for the session list, the list column, the toolbar and the search panel, so the design review shows the shipped label.

## What deliberately did **not** change

German prose such as *"eine Beratungsstelle in Ihrer Nähe"* is untouched. There "Nähe" means physical proximity, not the modality — renaming it would produce "eine Beratungsstelle in Ihrer Mail".

Internal identifiers (`nearby` chip id, `--nearby` CSS modifier, `data-cy="sessions-list-chip-nearby"`) are unchanged so Cypress selectors and styles keep working.

## ⚠️ Weblate

Translations are served from Weblate, which overrides the repo bundles. **This PR alone will not change the label on a deployed environment** — the same strings have to be updated in Weblate. The repo values matter for fresh imports and for the fallback path.

The non-German locales were set to `Mail` as well, since the modality is now a product term rather than a translated word. Translators can still override each of them in Weblate if a local term is preferred.

## Tests

`SessionsListToolbar.test.tsx` gained a case that renders the toolbar with no translations loaded and asserts the chip reads **Mail** and never **Nearby** — it covers the fallback path, which is what a missing Weblate string would hit. A new `modalityLabels.test.ts` pins the shipped label in de, de@informal and en.

`npx vitest run`: 258 files, 1833 tests green.

## Test plan

- [ ] Log in as a counsellor and open the session list: the filter chip row shows **Mail** instead of **Nähe**.
- [ ] Apply the **Mail** filter: only agency-counselling sessions remain.
- [ ] Open a session list item of that modality: the meta row next to the icon reads **Mail**.
- [ ] Open the search panel and the type filter: the entry reads **Mail**.
- [ ] Check the enquiry list chips (`Erstanfragen`): the chip reads **Mail**.
- [ ] Switch to English: all of the above read **Mail**.
- [ ] Registration flow: the postcode step still says *"Beratungsstelle in Ihrer Nähe"* — proximity wording must be intact.
- [ ] On mobile width (<768px): the chip row scrolls and **Mail** is not truncated.


Storypapst:
## ⚠️ Weblate blocks the visible effect

Reminder for whoever releases this: translations are served from Weblate, which overrides the repo bundles. Merging this PR alone will **not** change the label on pre-dev or dev. `sessionList.toolbar.chips.nearby` and `sessionList.toolbar.chips.chats` have to be set to `Mail` in Weblate as well.

What this PR does guarantee: the fallback path. `SessionsListToolbar.test.tsx` renders the toolbar with no translation loaded and asserts the chip reads **Mail** and never **Nearby** — so even a missing Weblate string can no longer surface the old wording.

kiodreambau:
## Repair pass — rebased on `pre-dev`, gates green locally

**What was broken.** Two separate things, neither of them the label change:

1. The branch was 160 commits behind `pre-dev`, so its checks were stale. The last PR-validation run did not fail on a test — it hit the runner's maximum execution time and was cancelled.
2. `pre-dev` itself is red on the #1101 catalogue guard. #1075 added `attachments.mediaCheck.error` to `de@informal` with a value identical to the formal catalogue, taking `collectRedundantOverlayKeys` from 596 to 597.

**What changed.** Merged `origin/pre-dev` (clean, no conflicts) and dropped that one redundant overlay value. The sparse informal overlay only carries strings that actually differ; i18next resolves the key through the `de@informal -> de` fallback, so the rendered text is unchanged. The same one-line fix is in #1105 — whichever lands first makes it a no-op for the other.

Most of this PR's content had already reached `pre-dev`; what remains net of the merge is the `fr`/`ru`/`ti`/`tr` chip labels plus the two tests.

**Verification** (node 22.23.2, the commands from `.github/actions/node-build`):

| Gate | Result |
|---|---|
| `npm run lint:scripts` (`eslint src --max-warnings=0 && tsc`) | exit 0 |
| `npm run test:unit` | **319 files, 3294 tests, 0 failures** |
| `CI=false npm run build` | exit 0 |

Shirloin (approved):

PR #1001, opened 2026-08-07
fix(nav,groupchat): single-person profile icon and no waiting-room timer in internal team chats (#981, #979)

### LINKED ISSUE #979: Unnecessary feature (timer) in internal team chats
> **Status · 11.08.2026** — **Fixed.** Frontend: PR #1001, merged into `pre-dev` (`de548e94`). Backend: ORISO-UserService#984 via PR #986 — internal team chats were being deleted an hour after being started. Needs deploy and retest.

---

### Affected Service/Component

Frontend (ORISO-Frontend)

### Bug Description

When a counsellor accesses the internal team chats in the conversations area and creates a new chat, the chat is currently treated as a start-based conversation flow.

The current implementation displays unnecessary start-chat information, a timer. Internal team chats should function as persistent chat rooms without a start flow, allowing asynchronous and synchronous communication between colleagues.

### Steps to Reproduce

1. Log in as a counsellor.
2. Navigate to the conversations area.
3. Open the internal team chats section.
4. Create a new internal team chat with one or multiple colleagues.
5. Observe that the created chat displays a start-chat flow with timer.

### Expected Behavior

Internal team chats should open directly as chat rooms without any timer, animation, or start-chat flow.

### Actual Behavior

When counsellors create an internal team chat, they see unnecessary start-chat information (a timer). 

### Screenshots/Videos

Image

https://github.com/user-attachments/assets/5ac35326-47f5-45f0-b6ad-a90380027234

### Acceptance Criteria
Internal team chats open directly as available chat rooms without requiring a start action, timer or other animations.

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before
- [x] I have provided all relevant information above
- [x] I can reproduce this bug consistently

### LINKED ISSUE #981: Incorrect profile icon displayed for counsellor accounts
> **Status · 11.08.2026** — **Fixed.** PR #1001, merged into `pre-dev` (`de548e94`). Needs deploy and retest on Pre-Dev.

---

### Affected Service/Component

Frontend (ORISO-Frontend)

### Priority

P1-High

### Bug Description

When a counsellor is logged in, the profile icon displayed in the user interface does not match the context of the counsellor profile.

The current icon represents a parent/child relationship, which is not relevant for a counsellor profile. The profile icon should represent a single user/person instead.

### Steps to Reproduce

1. Log in with a counsellor profile.
2. Navigate to the profile area **My profile** where the profile icon is displayed.

### Expected Behavior

In the "My profile" section, the logged-in counsellor profile should use a generic single-person icon that correctly represents an individual user profile.

### Actual Behavior

The counsellor profile is represented by an icon showing multiple persons (parent/child).

### Environment

_No response_

### Relevant Logs/Error Messages

```shell

```

### Screenshots/Videos

Image

https://github.com/user-attachments/assets/448a0d3b-faf9-49c3-a5d4-beafd9788a7f

### Workaround

_No response_

### Acceptance Criteria
The "My profile" section of a counsello displays a single-person profile icon.

- [x] I have searched existing issues to ensure this bug hasn't been reported before
- [x] I have provided all relevant information above
- [x] I can reproduce this bug consistently

Storypapst:
Closes #981
Closes #979

## #981 — the profile icon showed a parent and a child

The "My profile" item in the navigation rail used an inlined `supervised_user_circle` glyph: an adult figure with a smaller figure cut into the torso. Rendering the path data confirms it — the child is woven into the outline, so it could not be fixed by removing a subpath.

It now uses `profil_outline.svg` / `profil_filled.svg`, the design-system assets every sibling rail item already uses. The three inline `NavProfile*` components are removed.

**Note for the reviewer:** the icon is shared by the advice-seeker rail and the counsellor rail, so this changes both. That is intended — "My profile" is a single person for every role — but if the parent/child glyph was a deliberate choice for family counselling, say so and we will split it into two icons with a role condition.

## #979 — a countdown timer in internal team chats

Creating an internal team chat opened a waiting-room view with a countdown that immediately ran overdue and started counting up, with an animation.

The cause is a modality-blind predicate. UserService sets `active = false` on **every** V2 group chat (`CreateChatFacade.java:124`), so internal team chats land in the same branch as a scheduled self-help circle, and the start time is "now", which makes the countdown overdue on arrival.

The correct signal already existed and was simply never consulted at the branch point: `conversationType = INTERNAL_GROUP` is persisted correctly and `getModality()` reads it. New `getGroupChatWaitingAreaVisibility` helper, with tests, that suppresses the countdown, the calendar slot and the netiquette rules for `Modality.INTERNAL_GROUP`.

An author-written greeting is real content rather than waiting-room chrome, so it stays visible.

**Deliberately conservative:** the join/start button stays. The room still has to be opened server-side before messages can flow, so removing it would break the chat rather than tidy it.

## A separate, more serious finding — not fixed here

The workaround a counsellor would naturally reach for is destructive.

`DeactivateGroupChatService` deactivates every active chat once `startDate + duration` has passed. Internal chats ship `duration: 60`. `StopChatActionCommand:55-58` then calls `chatService.deleteChat(chat)` **and** `matrixChatShutdownService.shutdownRoom(chat)` for non-repetitive chats — and internal chats are `repetitive: false`.

So an internal team chat that is ever started is deleted, together with its Matrix room, roughly an hour later. That contradicts this issue's own acceptance criterion ("persistent chat rooms") and is a data-loss risk independent of the timer. It needs a backend change in ORISO-UserService — either create `INTERNAL_GROUP` chats with `active = true`, or exclude them from the deactivation sweep. Filed separately.

## Reviewer test plan

**#981**

- [ ] Log in as a counsellor, go to `/profile/allgemeines`
- [ ] The "Mein Profil" rail item shows a single-person glyph — zoom to 3x, the child figure is easy to miss at 24px
- [ ] Check all three states: inactive outline, hover, and active/filled
- [ ] The active state renders in the accent colour, not dark grey (a grey icon means the fill is not inheriting `currentColor`)
- [ ] Mobile viewport (≤900px), where the rail becomes the bottom bar
- [ ] Log in as an advice seeker and confirm the same item — this rail changes too
- [ ] Before/after screenshots attached to this PR

**#979**

- [ ] As counsellor A, create an internal team chat with colleague B
- [ ] Open it: the message stream renders directly — no countdown, no calendar slot, no netiquette rules, no animation
- [ ] Send a message and confirm B receives and can read it
- [ ] Log in as B and open the same chat: same result, no waiting area
- [ ] **Regression guard:** create a self-help circle with a future start. Open it as counsellor and as advice seeker — the waiting area **and** the countdown must still appear. This is the branch most at risk.
- [ ] A scheduled group chat that has already started still behaves as before
- [ ] `npm run test:unit -- groupChatHelpers` is green


Storypapst:
CodeRabbit hit its review limit before reaching this PR, so I reviewed it myself. No defect found, but one behavioural edge the reviewer should know about before testing on `dev`:

**Legacy group chats without a `conversationType` are classified as internal.** `getModality` falls back to a heuristic when the column is empty (`getModality.ts:94-97`): a group chat is `SELF_HELP` if it carries `repeatCount` or `repetitive`, otherwise `INTERNAL_GROUP`. So a legacy single-occurrence group chat with advice seekers — created before the column existed — would now lose its waiting area and countdown.

This is not reachable through the current create flows: `CreateConversationView` omits `repeatCount` for internal chats, and the circle settings view always sets it, so every chat created today classifies correctly. Only pre-column rows are affected.

Branching on the raw `conversationType` instead would avoid it, at the cost of bypassing the single modality selector that ADR-006 establishes. Given that there are no production users and that Pre-Dev data can be recreated, keeping the ADR-006 selector is the better trade — but it is worth knowing if an old chat on `dev` behaves unexpectedly during review.

The regression test in the plan ("create a self-help circle with a future start, confirm the waiting area still appears") is the one that matters most here.

Worth a re-review once the CodeRabbit limit resets — `@coderabbitai review`.

_🤖 Reviewed by [Claude Code](https://claude.com/claude-code)_

Storypapst:
@coderabbitai review

Issue #995, opened 2026-08-07
Heading menu: German labels missing and "Normaler Text" option absent


Storypapst:
**Status 2026-08-12 — delivered, awaiting test.** The heading menu is German and offers "Normaler Text" as the way back to body text. The whole `message.submit.toolbar.*` namespace now exists (40 keys), so every toolbar label is translated, not just the headings. PR #1007, merged to `pre-dev`. Next step: look at it on Pre-Dev once the environment is deployed.

---

Parent: #992
Affected repos: `OpenResilienceInitiative/ORISO-Frontend`

## Why

The composer's heading menu offers "Heading 1 … Heading 4" — untranslated developer language — and there is **no way back to normal body text** once a heading is applied. Counsellors write in German and do not think in H-levels. Reported by Frank, 2026-08-08.

## Where it was found

- `src/components/messageSubmitInterface/inputField/ComposerToolbar.tsx:52` — `const HEADING_LEVELS = [1, 2, 3, 4]`; the menu is built from that list only, so a paragraph entry does not exist.
- `ComposerToolbar.tsx:80-88` — labels come from `translate('message.submit.toolbar.heading{n}', 'Heading {n}')`.
- `src/resources/i18n/de/common.json` and `.../en/common.json` — **the entire `message.submit.toolbar.*` namespace is missing in both files**, so every toolbar label (bold, italic, quote, link, list, …), not just the headings, renders from its hardcoded English fallback.
- `src/components/messageSubmitInterface/TipTapComposer.tsx` — the heading actions map to `toggleHeading({ level })`; there is no `setParagraph` action.

## What to build

- [ ] Add a **"Normaler Text"** entry as the first item of the heading menu, wired to `setParagraph()`, selected whenever the caret is in a plain paragraph.
- [ ] Rename the four heading entries to the agreed German wording:

| Level | German label |
| --- | --- |
| paragraph | Normaler Text |
| h1 | Titel |
| h2 | Große Überschrift |
| h3 | Mittlere Überschrift |
| h4 | Kleine Überschrift |

- [ ] Add the whole missing `message.submit.toolbar.*` namespace to `de` and `en` (headings **and** the remaining toolbar labels that currently fall back to English), plus `message.mention.notInChat`.
- [ ] Push the new keys to Weblate — Weblate overrides the repo files at runtime, so repo-only keys will not reach the deployed app.
[...]

Issue #993, opened 2026-08-07
@-mention picker cannot mention a counsellor


Storypapst:
**Status 2026-08-12 — not fixed.** PR #1006 is merged, but it only makes the failure visible: the picker now says whether it is loading, whether nobody was found, or whether the list could not be loaded. Before, all three collapsed into the same empty list and the error was swallowed twice. Next step is a check on Pre-Dev — type `@` in a chat and read which of the three messages appears. That tells us whether the request fails, returns nothing, or the counsellor is filtered out.

---

Parent: #992
Affected repos: `OpenResilienceInitiative/ORISO-Frontend`

## Why

Typing `@` in the chat composer opens the mention popup, but no counsellor can be selected — the list stays empty, so intentional mentions cannot be used at all. Reported by Frank, 2026-08-08.

## Where it was found

- `src/components/messageSubmitInterface/messageSubmitInterfaceComponent.tsx:2140-2215` — the effect that loads `agencyConsultantDirectory` from `apiGetAgencyConsultantList(agencyId)`
- `src/components/messageSubmitInterface/messageSubmitInterfaceComponent.tsx:3306-3380` — `mentionDataRef` / `mentionProvider`
- `src/components/messageSubmitInterface/inputField/extensions/createMentionExtension.tsx`
- `src/components/messageSubmitInterface/inputField/extensions/mentionFiltering.ts`

## Suspected causes (to be confirmed on Pre-Dev, not yet proven)

1. **The directory silently ends up empty.** Every failure path — asker-only or anonymous chat, missing `agencyId` on the active session, or *any* rejected API call — resolves to `setAgencyConsultantDirectory(new Map())` with no error surfaced (`.catch(() => …)` at line 2206). An empty directory produces an empty popup that looks identical to "no match".
2. **Self-exclusion may over-match or under-match.** `filterMentionCandidates` drops `candidate.id === selfId`, comparing a `consultantId` from the agency list against `userData.userId`. If those id spaces differ, the wrong row is dropped or the author is offered.
3. **The provider closure is captured once on mount** (`TipTapComposer.tsx:223-227`) and reads through `mentionDataRef`; if the directory resolves after the extension is built and the ref update does not land, `getCandidates()` returns the stale empty snapshot.

Confirm which one it is before fixing — do not fix all three blind.

## What to build

[...]

PR #1006, opened 2026-08-07
Make the empty @-mention picker say why it is empty (#993)

Storypapst:
Refs OpenResilienceInitiative/ORISO-Frontend#993

> **Base branch:** this PR targets `fix/995-composer-heading-and-toolbar-i18n`, not `pre-dev`, because it consumes the `message.mention.*` strings that PR adds. Merge #995 first; GitHub will retarget this one to `pre-dev` automatically.

> **This does not close the issue.** The root cause is not confirmed yet. What this lands is the ability to find it — see below.

## The problem with the old code

Every failure path collapsed into the same empty `Map`, and the popup rendered *literally nothing* when the list was empty. A rejected request, a session without an `agencyId`, and an agency with no other consultants were indistinguishable — to the user and to us. The user just saw the `@` do nothing.

The error was swallowed twice: once in `apiGetAgencyConsultantList` (`catch { return [] }`) and again in the composer (`.catch(() => setAgencyConsultantDirectory(new Map()))`).

## What changed

- **`fetchAgencyConsultantList`**, a throwing variant, so a caller can tell "no consultants" from "request rejected". The swallowing helper stays for the callers that only want a list.
- The composer tracks the directory as **loading / ready / error / unavailable**, and reports a failed or `agencyId`-less load through `apiPostError` at WARN.
- **The popup now says why it is empty:** "Liste konnte nicht geladen werden", "Niemand gefunden", or a loading line. It stays silent only where mentions do not apply at all (asker, anonymous chat).

## What I could not confirm

I could not reproduce this against live data, so I am not claiming a fix. One lead worth checking first: `/users/consultants` requires the `VIEW_AGENCY_CONSULTANTS` authority. The plain `CONSULTANT` role does carry it — I checked `Authority.java` — but **`GROUP_CHAT_CONSULTANT` does not**, so a group-chat consultant would get a 403 and, before this PR, a silent blank popup.

With this merged, reproducing on Pre-Dev with the network tab open should name the cause in one attempt.

## How it was verified

- `npx vitest run` — 261 files, 1871 tests, all passing
- 9 new tests: `MentionList.emptyStates.test.tsx` (each cause gets its own message) and `apiGetAgencyConsultantList.test.ts` (the throwing variant really throws)
- `tsc`, Storybook `tsc`, `eslint --max-warnings=0`, `stylelint` clean

## Reviewer test plan

- [ ] As a counsellor, open a 1-1 conversation and type `@` → either a list of counsellors appears, or a line saying why it is empty — never a silent nothing
- [ ] Block `/users/consultants` in the browser devtools and type `@` → "Liste konnte nicht geladen werden" appears and a WARN reaches the error endpoint
- [ ] Type `@` followed by nonsense that matches nobody → "Niemand gefunden"
- [ ] As an advice seeker in an anonymous chat, type `@` → nothing appears (this is intended, mentions are a counsellor tool)
- [ ] Repeat the first check on a 320px and a 412px wide phone → the popup stays on screen and does not cover the send button
- [ ] Switch the UI to German and to English → the status lines are translated in both
- [ ] **If the list is still empty in a normal counselling chat**, note your Keycloak roles and the `/users/consultants` response code on the issue — that is the missing piece

### DIFF
--- a/src/api/apiGetAgencyConsultantList.test.ts
+++ b/src/api/apiGetAgencyConsultantList.test.ts
@@ -1,0 +1,52 @@
+/**
+ * #993 — the consultant list was swallowed twice: once here, once in the
+ * composer. A 403 therefore looked exactly like "this agency has no other
+ * consultants". The throwing variant exists so a caller can tell them apart.
+ */
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('../resources/scripts/endpoints', () => ({
+	endpoints: { agencyConsultants: '/service/users/consultants' }
+}));
+
+const fetchDataMock = vi.fn();
+vi.mock('./fetchData', () => ({
+	fetchData: (...args: unknown[]) => fetchDataMock(...args),
+	FETCH_METHODS: { GET: 'GET' },
+	FETCH_ERRORS: { CATCH_ALL: 'CATCH_ALL' }
+}));
+
+const importModule = async () => await import('./apiGetAgencyConsultantList');
+
+afterEach(() => {
+	fetchDataMock.mockReset();
+});
+
+describe('agency consultant list', () => {
+	it('asks for the consultants of the given agency', async () => {
+		fetchDataMock.mockResolvedValue([]);
+		const { fetchAgencyConsultantList } = await importModule();
+
+		await fetchAgencyConsultantList('42');
+
+		expect(fetchDataMock).toHaveBeenCalledWith(
+			expect.objectContaining({
+				url: '/service/users/consultants?agencyId=42'
+			})
+		);
+	});
+
+	it('lets the caller see a rejected request', async () => {
+		fetchDataMock.mockRejectedValue(new Error('403'));
+		const { fetchAgencyConsultantList } = await importModule();
+
+		await expect(fetchAgencyConsultantList('42')).rejects.toThrow('403');
+	});
+
+	it('keeps the swallowing variant for callers that only want a list', async () => {
+		fetchDataMock.mockRejectedValue(new Error('403'));
+		const { apiGetAgencyConsultantList } = await importModule();
+
+		await expect(apiGetAgencyConsultantList('42')).resolves.toEqual([]);
+	});
+});
--- a/src/api/apiGetAgencyConsultantList.ts
+++ b/src/api/apiGetAgencyConsultantList.ts
@@ -10,17 +10,30 @@ export interface Consultant {
 	isSupervisor?: boolean;
 }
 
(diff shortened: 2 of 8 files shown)


Storypapst:
## Verification evidence

|---|---|---|---|
| INFORMATIONAL | Local | `0793b4f` | Claude |

### Storybook, German UI\. Before this change an empty popup rendered nothing at all\. Now: a failed consultant\-list request says so in the error colour, a query matching nobody says Niemand gefunden, and a populated list is unchanged\. INFORMATIONAL, not PASS \- this proves the diagnosis is visible, not that the underlying bug is fixed; the root cause still needs one Pre\-Dev reproduction\.

Storybook, German UI\. Before this change an empty popup rendered nothing at all\. Now: a failed consultant\-list request says so in the error colour, a query matching nobody says Niemand gefunden, and a populated list is unchanged\. INFORMATIONAL, not PASS \- this proves the diagnosis is visible, not that the underlying bug is fixed; the root cause still needs one Pre\-Dev reproduction\.

### Storybook, German UI\. Before this change an empty popup rendered nothing at all\. Now: a failed consultant\-list request says so in the error colour, a query matching nobody says Niemand gefunden, and a populated list is unchanged\. INFORMATIONAL, not PASS \- this proves the diagnosis is visible, not that the underlying bug is fixed; the root cause still needs one Pre\-Dev reproduction\.

Storybook, German UI\. Before this change an empty popup rendered nothing at all\. Now: a failed consultant\-list request says so in the error colour, a query matching nobody says Niemand gefunden, and a populated list is unchanged\. INFORMATIONAL, not PASS \- this proves the diagnosis is visible, not that the underlying bug is fixed; the root cause still needs one Pre\-Dev reproduction\.

### Storybook, German UI\. Before this change an empty popup rendered nothing at all\. Now: a failed consultant\-list request says so in the error colour, a query matching nobody says Niemand gefunden, and a populated list is unchanged\. INFORMATIONAL, not PASS \- this proves the diagnosis is visible, not that the underlying bug is fixed; the root cause still needs one Pre\-Dev reproduction\.

Storybook, German UI\. Before this change an empty popup rendered nothing at all\. Now: a failed consultant\-list request says so in the error colour, a query matching nobody says Niemand gefunden, and a populated list is unchanged\. INFORMATIONAL, not PASS \- this proves the diagnosis is visible, not that the underlying bug is fixed; the root cause still needs one Pre\-Dev reproduction\.

[All evidence for this run](https://evidence.dreambau.com/r/sifaegbdtqy6edvsczunau7y28i3dku4)

shanzaeimran (approved):

Issue #996, opened 2026-08-07
Replace the voice-note player and show an insertion marker in the composer


Storypapst:
**Status 2026-08-12 — delivered, awaiting test.** The designed waveform player replaces the placeholder, and the composer shows an insertion marker at the caret. PR #1004, merged to `pre-dev`. Next step: look at it on Pre-Dev once the environment is deployed.

---

Parent: #992
Affected repos: `OpenResilienceInitiative/ORISO-Frontend`

## Why

The voice-note player in the chat is still the placeholder that was written to unblock audio playback: a text glyph button (`▶` / `||`), a single filled `div` as the "waveform", and a duration label. It is not the designed player, it is not tokenized, and it is not accessible. It has to be replaced by the designed waveform player. Second part: when a voice note is inserted from the editor, the composer must show **where** it will be placed — a visible insertion point at the caret. Requested by Frank, 2026-08-08.

Design source (local, not in the repo): `ORISO/Claude Design/Voice media Player/Media Player/` — `VoicePlayer.dc.html`, `Waveform.dc.html`, `Chat message polish.dc.html` (voice-message options 1d/1e). Attach the exported screens to this issue before implementation starts.

## Where it was found

- `src/components/message/MessageAttachment.tsx:262-320` — `renderVoiceAttachment`: text-glyph play button, `messageItem__voiceNote__wave` / `__wavePlayed` progress `div`, raw `<audio>` element.
- `src/components/message/message.styles.scss`, `messageVoiceNote.styles.test.ts`
- Composer side: `src/components/messageSubmitInterface/messageSubmitInterfaceComponent.tsx` (recording/upload path), `src/components/messageSubmitInterface/TipTapComposer.tsx` (insertion point).

## What to build

**A — Replace the player**

- [ ] Build a reusable `VoicePlayer` component from the design: real play/pause icon button, waveform bars with played/unplayed state, elapsed / total time, seek by click or drag on the waveform, and playback speed if the design specifies it.
- [ ] Style it exclusively with `--m3-*` tokens — no hardcoded greens or off-system colors (this also settles the voice-player half of #568).
- [ ] Use it in all voice-note surfaces: plain attachment, encrypted attachment after decryption, and the composer preview of a recording.
- [ ] Accessibility: the play button is a real `<button>` with an accessible name, the waveform exposes a slider role with keyboard seeking, and state changes are announced.
[...]

Issue #994, opened 2026-08-07
Attachment download card is off the design system


Storypapst:
**Status 2026-08-12 — delivered, awaiting test.** The attachment card is now a tokenized, class-based card in all four states (plain, encrypted, unchecked, blocked); no inline styles remain. PR #1005, merged to `pre-dev`. Next step: look at it on Pre-Dev once the environment is deployed.

---

Parent: #992
Affected repos: `OpenResilienceInitiative/ORISO-Frontend`

## Why

The download / file-attachment card that a message shows after a file is sent from the editor looks unfinished and does not match the rest of the chat surface. Reported by Frank, 2026-08-08 ("Download hat komisches Design").

> Scope check for the reviewer: this is written against the attachment/download card in the chat (`MessageAttachment`). If a different download surface was meant, say so on this issue and it will be retargeted before work starts.

## Where it was found

`src/components/message/MessageAttachment.tsx` — the non-encrypted branch renders the download card as a bare `<a download>` carrying **inline styles** (`textDecoration`, `color`, `cursor`) instead of the class-based, tokenized card the rest of the chat uses. The encrypted branch and the media-check states (`--unchecked`, blocked) each build their own variant, so the same file shows up in four visually different shapes depending on encryption and scan state.

Related styles: `src/components/message/message.styles.scss`.

## What to build

- [ ] Rebuild the attachment card as one component with a single visual definition, used by all four states (plain, encrypted/locked, awaiting media check, blocked).
- [ ] Remove the inline `style={{…}}` props; move everything into the SCSS using `--m3-*` tokens — no hardcoded colors, radii or spacing.
- [ ] Make the download affordance explicit: file-type icon, file name (truncating on one line, with a title attribute), type and size line, and a visible download icon/button so it does not read as plain text.
- [ ] Give it real hover, focus-visible, and active states; the whole card must be reachable and triggerable by keyboard.
- [ ] Cover the states in Storybook next to the existing `MessageAttachment.stories.ts`.

## Acceptance

- [ ] All four attachment states share one card design and differ only by their state affordance.
- [ ] No inline style props and no non-token colors remain in `MessageAttachment.tsx`.
- [ ] A long file name truncates instead of breaking the bubble layout.
- [ ] The card has a visible focus ring and is operable with the keyboard.
- [ ] Light and dark scheme both render correctly.

## Reviewer test plan

[...]

Issue #568, opened 2026-07-20
Align mandatory 2FA and voice-player UI with M3 tokens


Storypapst:
## Why

Three small shared Frontend UI defects add avoidable friction or violate the M3 token system: mandatory 2FA starts with a redundant nag, the setup stepper becomes unreadable on mobile, and the voice player uses unrelated green colors.

## Problem

The mandatory security flow and reusable voice-message control are not using the intended direct interaction and design-token behavior.

## Actual behavior

- Mandatory 2FA first shows an extra explanatory dialog before the actual setup.
- The five-step setup indicator collides and is hard to read on narrow screens.
- The voice player uses hardcoded green styling instead of the active M3 primary palette and its tonal roles.

## Expected behavior

- Mandatory 2FA opens the setup dialog directly while the optional reminder behavior remains available where applicable.
- The setup steps remain readable on mobile.
- The voice player derives its background, controls, progress, and contrast from M3 primary token roles.

## What to build

- Route mandatory 2FA directly into the existing setup dialog.
- Make the setup stepper responsive without removing information.
- Replace the voice-player green palette with shared primary token roles.
- Add focused component/style regression tests.

## Acceptance

- [x] Mandatory 2FA has no redundant intermediate click.
- [x] The complete setup flow remains readable at 390 x 844.
- [x] The voice player uses M3 primary token roles rather than hardcoded green.
- [x] Focused tests and desktop/mobile browser evidence exist.

## Evidence

Run: `e2e-20260719-1711`

- Direct 2FA desktop/mobile: `screenshots/08-two-factor-direct-local-desktop.png`, `screenshots/08-two-factor-direct-local-mobile.png`
- Stable setup desktop/mobile: `screenshots/08-two-factor-dialog-stable-local-desktop.png`, `screenshots/08-two-factor-dialog-stable-local-mobile.png`
- Failed 2FA attempt: `screenshots/08-two-factor-direct-local-failed.png`
- M3 voice player desktop/mobile: `screenshots/20-voice-player-m3-local-desktop.png`, `screenshots/20-voice-player-m3-local-mobile.png`
- Failed player attempt: `screenshots/20-voice-player-m3-local-failed.png`

Storypapst:
The voice-player part of this issue is superseded by #996 (sub-issue of #992), which replaces the placeholder player in `MessageAttachment.tsx` with the designed waveform player on M3 tokens. The 2FA parts of this issue are unaffected.

[...]

PR #1007, opened 2026-08-07
Designed voice player + composer insertion marker (#996)

Storypapst:
Closes OpenResilienceInitiative/ORISO-Frontend#996

> **Base branch:** this PR targets `fix/994-attachment-download-card`, not `pre-dev`, because both change `MessageAttachment`. Merge #994 first; GitHub will retarget this one automatically.

## Part A — the real player

The chat still shipped the stopgap control: a `▶`/`||` **text glyph**, a single filled `div` as the "waveform", and a raw `<audio>` element.

This picks Frank's existing `VoicePlayer` work off `feat/voice-player-design` and rebases it onto `pre-dev` (that branch was 88 commits behind and had no PR).

- Two variants: **tiles** (one per second up to 16, above that 16 sections with the seconds per tile in the meta line) and **ring** for narrow rows. Both seek, and both run off a real audio element rather than a timer.
- Used for plain and for decrypted voice notes, and now also for the **composer's preview** — what you check before sending looks like what the other side receives.
- `--m3-*` tokens throughout, which also settles the voice-player half of #568.

**Deliberately not picked** from the source commit: the encryption-notice redesign (`E2EEActivatedMessage`, its SCSS, the `e2ee-hint` variables, the `e2ee.notice` strings). That is a separate change and does not belong in this issue. The colour guard that shipped with it was narrowed to the voice player for the same reason.

## Part B — the insertion marker

Recording types nothing into the editor, so there was no way to tell where in a half-written message the recording would land.

`InsertionMarker` is a ProseMirror **widget decoration** pinned at the caret the moment recording starts. It never enters the document and never reaches the transport markup, and its position maps through every transaction, so it follows the text if the user keeps typing around it.

**One deviation from the issue's acceptance list:** the issue said the marker must not survive a blur. In practice pressing the mic button blurs the editor, so a blur-clear would wipe the dot at the exact moment it becomes useful. The marker instead lives for as long as the insertion is pending — through the blur, through the preview — and a content reset (send, clear, draft switch) drops it. Honours `prefers-reduced-motion`.

## How it was verified

- `npx vitest run` — 261 files, 1871 tests, all passing
- 6 new tests in `insertionMarker.test.ts`, including one asserting the dot never becomes document content
- 26 tests came with the picked VoicePlayer work and still pass
- `tsc`, Storybook `tsc`, `eslint --max-warnings=0`, `stylelint` clean
- New Storybook story `Components/Composer/InsertionMarker` so the marker is reviewable without granting microphone access

## Reviewer test plan

- [ ] Record and send a voice note → the sent message shows the new player with a waveform, not a `▶` character
- [ ] Press play, then pause → the icon changes, the played part of the waveform fills, the elapsed time counts up and stops
- [ ] Click into the middle of the waveform → playback jumps to that position and the time label matches
- [ ] Receive a voice note in an encrypted conversation → after decryption the same player appears and plays
- [ ] Tab to the player and operate it with the keyboard only → it works and the focus ring is visible
- [ ] Start a voice note with the caret in the middle of a written sentence → a dot shows exactly where the voice note will be inserted, and it stays while you press Stop
- [ ] Cancel the recording → the dot disappears and the text is unchanged
- [ ] Keep typing before the dot while a recording is pending → the dot stays with the text it was placed in
- [ ] Repeat the record → preview → send round trip on a 320px and a 412px wide phone by touch → the waveform is seekable by dragging and nothing overflows
- [ ] Switch to the dark scheme → waveform, icons and time label stay readable


Storypapst:
## Verification evidence

|---|---|---|---|
| PASS | Local | `6ca0519` | Claude |

### Storybook, German UI\. Voice player: tiles variant in a conversation \(own message pink, incoming grey\) and all variants side by side\. Insertion marker: the red dot marks where a pending voice message will land; second shot is the same story with nothing pending\.

Storybook, German UI\. Voice player: tiles variant in a conversation \(own message pink, incoming grey\) and all variants side by side\. Insertion marker: the red dot marks where a pending voice message will land; second shot is the same story with nothing pending\.

### Storybook, German UI\. Voice player: tiles variant in a conversation \(own message pink, incoming grey\) and all variants side by side\. Insertion marker: the red dot marks where a pending voice message will land; second shot is the same story with nothing pending\.

Storybook, German UI\. Voice player: tiles variant in a conversation \(own message pink, incoming grey\) and all variants side by side\. Insertion marker: the red dot marks where a pending voice message will land; second shot is the same story with nothing pending\.

### Storybook, German UI\. Voice player: tiles variant in a conversation \(own message pink, incoming grey\) and all variants side by side\. Insertion marker: the red dot marks where a pending voice message will land; second shot is the same story with nothing pending\.

Storybook, German UI\. Voice player: tiles variant in a conversation \(own message pink, incoming grey\) and all variants side by side\. Insertion marker: the red dot marks where a pending voice message will land; second shot is the same story with nothing pending\.

### Storybook, German UI\. Voice player: tiles variant in a conversation \(own message pink, incoming grey\) and all variants side by side\. Insertion marker: the red dot marks where a pending voice message will land; second shot is the same story with nothing pending\.

Storybook, German UI\. Voice player: tiles variant in a conversation \(own message pink, incoming grey\) and all variants side by side\. Insertion marker: the red dot marks where a pending voice message will land; second shot is the same story with nothing pending\.

[All evidence for this run](https://evidence.dreambau.com/r/th64577vi3qxj6jmb774k4232r4cdt6p)

shanzaeimran (approved):

